PR #8, opened 2017-09-26
Add the vagrantfile and key folder

petrkotas:
@cynepco3hahue Hi I have added my vagrant file with small tweaks. I will continue to integrate the vagrant even more.

PRM


petrkotas:
Thanks, I would merge it too :-)

PR #9, opened 2017-10-05
Add config file for Vagrantfile

petrkotas:
To be more flexible and more user friendly
this patch adds the config file used
to configure the dev environment.

Originally the env variables were used,
but this is not a transparent way of
configuration.

@cynepco3hahue if you do not have objection please review and merge.


cynepco3hahue:
Looks good for me

PR #12, opened 2017-10-19
deploy-kubernetes: replace golang role, we removed it past

lukas-bednar:
kubevirt-deploy.yml playbook fails on absence of golang role.

### DIFF
--- a/deploy-kubernetes.yml
+++ b/deploy-kubernetes.yml
@@ -2,11 +2,15 @@
 - hosts: all
   remote_user: root
   vars_files:
-    - "{{ playbook_dir }}/kubernetes/vars/default_vars.yml"
-    - "{{ playbook_dir }}/common/vars/default_vars.yml"
+  pre_tasks:
+    - name: override roles variables
+      include_vars: "{{ item }}"
+      with_items:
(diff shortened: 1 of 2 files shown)


deploy-kubernetes.yml
```diff
@@ -2,11 +2,15 @@
...
-    - "{{ playbook_dir }}/common/vars/default_vars.yml"
+  pre_tasks:
+    - name: override roles variables
+      include_vars: "{{ item }}"
+      with_items:
+        - "{{ playbook_dir }}/kubernetes/vars/default_vars.yml"
+        - "{{ playbook_dir }}/common/vars/default_vars.yml"
   roles:
     - "{{ playbook_dir }}/kubernetes/roles/prerequisites"
-    - "{{ playbook_dir }}/common/roles/golang"
```
fabiand: Is there also a fix upstream against this role?

lukas-bednar: This role was removed from this repository by https://github.com/cynepco3hahue/kubevirt-ansible/pull/9/files
But unfortunately it wasn't removed from `deploy-kubernetes.yml` file.

Does it answer your Q? If it is not let me know :-)

PR #19, opened 2017-10-30
kubevirt-common: replace glide with dep

lukas-bednar:
follow up with recent changes in kubevirt repository


lukas-bednar:
Verified on my environment

PR #34, opened 2018-01-07
automation: Fix log collection

gbenhaim:
- Fix the quoting of the trap.
  The current implementation caused "$run_path" to be evaluated to
  the empty string, because it occurred outside of the main function.

- Don't use "find" to copy "ansible.log", since it can cause a race where
  "find" copies "ansible.log" and then finds it again in the exported
  artifacts dir.


gbenhaim:
ci test please

PR #35, opened 2018-01-07
deploy-with-lago: Add more storage

gbenhaim:
- Add 10GB (sparse) disk and mount it to /var/lib/docker (needed when building
  Kubevirt's containers).

- Mount 100GB disk (sparse) in /opt in the master's VM. This disk will be
  used for the nfs share.

- Set "docker_dev" variable which marks "docker-storage-setup" role
  which disk to use.


fabiand:
lgtm

PR #49, opened 2018-02-01
untie playbooks

gbenhaim:
Adding control.yml which includes all the steps required
for installing and testing kubevirt.

With this change, each playbook can be used on its own.


fabiand:
Please have at leats on eperson revieweing stuff.

PR #72, opened 2018-02-16
Unite name of parameter for kubeconfig in repository

lukas-bednar:
We have roles for Kubernetes which use 'kubeconfig' parameter.
And next to it we have roles for OpenShift and here we use 'kconfig' as a
name for kubeconfig parameter.

This change unite name of this parameter to be 'kubeconfig' for all
roles in this repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ playbook to install KubeVirt from given manifest.
 | ---------------- | ------------- | -------------------------------------- |
 | kubevirt\_mf     | string | Path to KubeVirt manifest, you can get one from [releases](https://github.com/kubevirt/kubevirt/releases) or build it from sources |
 | openshift\_ansible\_dir | string | Path to [OpenShift Ansible project][openshift-ansible-project] |
-| kubeconfig | string | Path to kubeconfig |
+| kubeconfig | string | Path to the [`~/.kube/config`](https://docs.openshift.com/container-platform/3.7/cli_reference/manage_cli_profiles.html#switching-between-cli-profiles) file containing authentication information for the desired cluster |
 
 ```bash
 $ # Get KubeVirt manifest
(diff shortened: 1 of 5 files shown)


lukas-bednar:
The check is failing on missing dependency of `openshift_facts` I have a fix for it here #73

README.md
```diff
@@ -146,14 +146,14 @@ playbook to install KubeVirt from given manifest.
 | ---------------- | ------------- | -------------------------------------- |
 | kubevirt\_mf     | string | Path to KubeVirt manifest, you can get one from [releases](https://github.com/kubevirt/kubevirt/releases) or build it from sources |
 | openshift\_ansible\_dir | string | Path to [OpenShift Ansible project][openshift-ansible-project] |
-| kconfig | string | Path to kubeconfig |
+| kubeconfig | string | Path to kubeconfig |
```
alexxa: @lukas-bednar Please elaborate where kubeconfig, a KubeVirt(?) configuration file, comes from. Then one won't question it and will not search for it.

lukas-bednar: Addressed

README.md
```diff
@@ -146,7 +146,7 @@ playbook to install KubeVirt from given manifest.
 | ---------------- | ------------- | -------------------------------------- |
 | kubevirt\_mf     | string | Path to KubeVirt manifest, you can get one from [releases](https://github.com/kubevirt/kubevirt/releases) or build it from sources |
 | openshift\_ansible\_dir | string | Path to [OpenShift Ansible project][openshift-ansible-project] |
-| kubeconfig | string | Path to kubeconfig |
+| kubeconfig | string | Path to kubeconfig containing authentication information for desired cluster. |
```
alexxa: @lukas-bednar I still don't follow where kubeconfig comes from. Where does one can find that file? And I would replace `kubeconfig` in description (that's why it's called description) with `the KubeVirt|Kubernetes(?) configuration file`.

gbenhaim: @alexxa https://docs.openshift.com/container-platform/3.7/cli_reference/manage_cli_profiles.html

alexxa: @gbenhaim Good! So, I would put it as: 

Path to the [`~/.kube/config`](https://docs.openshift.com/container-platform/3.7/cli_reference/manage_cli_profiles.html#switching-between-cli-profiles) file containing authentication information for desired cluster

alexxa: @lukas-bednar ^^

lukas-bednar: Addressed

PR #74, opened 2018-02-16
Update CONTRIBUTING.md

alexxa:
* There are good Contributing templates and guidelines how to write it.
* I think contributing should be about contributing only. Anth else, for example, notes on repo structure and project organization can go into README.md in a root directory. The content of the current README, i.e. instructions how to use playbooks should belong to the playbooks subdirectory after repo reorganization.
* I wouldn't hardcore maintainers or any other names and contact details. It can become outdated any moment and  it requires extra maintenance. The latest info on all contributors can be extracted from Git Repo Info or by Git commands.
* What about not merging TODOs on master, but opening issues instead? I have removed all TODOs from the changed file and will duplicate them accordingly in a new issue if this PR is merged.
* When using abbreviations, please be sure those are defined first. For example, I don't see APB and K8S were defined. I'm new on a project, it took me a while to understand it. One shouldn't question what it is.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,98 +1,96 @@
-# Contributing to KubeVirt-Ansible
+# Contributing to KubeVirt Ansible
 
-## Intro
+Thank you for interest in contributing to KubeVirt Ansible! :tada::+1:
 
-In addition for allowing deployment of KubeVirt via Ansible,
-the KubeVirt-Ansible project is open for any contributions from
-Additional projects which aims to integrate with KubeVirt, such 
-as a specific storage or network project.
+The following is a set of guidelines for contributing to this project. These are not rules. Use your best judgment, and feel free to propose changes to this document in a pull request. Please make sure you are welcoming and friendly in all of our spaces.
 
-The integration will be in the form of Ansible roles per component
-which might later be integrated into a main Ansible Playbook Bundle.
+## Questions ? Help ? Ideas ? :bulb:
 
-Information on how to contribute will follow in the next sections.
+Please, don't use the issue tracker for support questions. Stop by the [#kubevirt](https://webchat.freenode.net/?channels=kubevirt) chat channel on freenode IRC or contact us on the [KubeVirt Google Group](https://groups.google.com/forum/#!forum/kubevirt-dev).
 
-## KubeVirt Ansible repository structure and roles
+## Reporting Bugs :bug:
 
-This section will describe the current repository structure and
-give a short example of an existing role for KubeVirt.
+Before creating bug reports, please check a [list of known issues](https://github.com/kubevirt/kubevirt-ansible/issues) to see if the problem has already been reported.
 
-### Roles Structure
+If not, go ahead and [make one](https://github.com/kubevirt/kubevirt-ansible/issues/new)! Be sure to include a **descriptive title and clear description** and please include **as many details as possible** to help maintainers reproduce the issue and resolve it faster. If possible, add a **code sample** or an **executable test case** demonstrating the expected behavior that is not occurring.
 
-[TODO]: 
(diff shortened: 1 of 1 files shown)


eedri:
Hi, thanks for helping to improve the document! 
Some comments on your points and I'll also review the contribution itself later on ( its long :) 

> I think contributing should be about contributing only. Anth else, for example, notes on repo structure and project organization can go into README.md in a root directory. The content of the current README, i.e. instructions how to use playbooks should belong to the playbooks subdirectory after repo reorganization.

True, however, some people might go into the contributing doc and get lost if they don't see any existing examples or reference for roles or code. I got comments from @rthallisey on it as well.
My intention was to add a link to the README file with the relevant info, rather than copy pasting the same info from there. 
This also applies for other sections like contributing tests, which can have its own file and we'll just add a link to it from the contributing file.

> I wouldn't hardcore maintainers or any other names and contact details. It can become outdated any moment and it requires extra maintenance. The latest info on all contributors can be extracted from Git Repo Info or by Git commands.

This was actually taken from the KubeVirt contributing file, It doesn't mean its the right way to go, but I agree to some degree. Its true that things might get outdated and maintainers are changing in time, however, due to the stage the project is in, and the needs to easy contributing I think specifying POC while the project is being in the early stage of getting contributing from outside makes sense. 

> What about not merging TODOs on master, but opening issues instead? I have removed all TODOs from the changed file and will duplicate them accordingly in a new issue if this PR is merged.

I think having new ISSUES makes sense, however, I didn't open them yet ( I have them on a Trello board ) because some are blocked on other issues and some are pending assignments to people.
My only concern with removing the TODO's is that we'll lose the intended place of what we want to add to the DOC and visibility of what is missing when looking at the current doc.

> When using abbreviations, please be sure those are defined first. For example, I don't see APB and K8S were defined. I'm new on a project, it took me a while to understand it. One shouldn't question what it is.

These are not unique to the project, its common components of the ecosystem we're in, coming from Kubernetes and Ansible.  
I agree it would help to define those somewhere, but I'm not sure it should be in any doc they are mentioned, maybe we should add links at the end for those? another options is to use the full name on the first occurance and then move to the abbreviations

CONTRIBUTING.md
```diff
@@ -1,98 +1,98 @@
...
+  * Please ensure all code changes reflected in documentation accordingly.
+  
+For me codding guidelines please see [Additional Links](#additional-links).
+
+### KubeVirt Ansible Contributors
+
+To see the list of KubeVirt Ansible Contributors, please
+
+* Check the [Contributors](https://github.com/kubevirt/kubevirt-ansible/graphs/contributors) page or
+* Run `git log --all --format='%aN <%cE>' | sort -u` in a cloned [KubeVirt Ansible](https://github.com/kubevirt/kubevirt-ansible) git repository.
```
fabiand: Or: `git shortlog -sne`

eedri:
In Gerrit we have 'draft' for this ( skipping CI ), is there a similar feature in GitHub for WIP PRs?

gbenhaim:
@eedri I think that the intention was to skip CI when only the docs were changed.

eedri:
Yes, I understood, I think we can automatically do it inside check-patch.sh or in the automation.yaml ( i.e add 'if code = text' skip tests ).. and then you don't have to handle new tags or add something to the header.

alexxa:
@eedri, @gbenhaim  what if text contains samples of code? will `if code=text skip tests` pass it?

eedri:
if code = text as just psuedo code example, in reality we can make it very fine grained like if certain files were updated, e.g via regex:

If change is in 'contributing.md, readme.md' or *.md, something like that..

gbenhaim:
@alexxa Can you please squash the commits into one (and do a rebase)?

Issue #82, opened 2018-02-20
Use std-ci yum repos mirrors


gbenhaim:
- It will protect the job from failure when Cento's repos are not available.
- It will accelerate the job (the mirrors and the Jenkins slaves are in the same data center).

Issue #44, opened 2018-01-25
Accelerate CI


gbenhaim:
- [ ] Cache Openshift containers on the Jenkins slaves
- [x] Run test suite in parallel.
- [x] Don't run full system update, update just the required packages.
- [ ] Install only needed deps, for example when running `install-kubevirt-release.yml`, there is no need to install `andrewrothstein.go-dev` from ansible galaxy.
- [x] Add a profiler to Ansible
- [x] Use std-ci yum repos mirrors https://github.com/kubevirt/kubevirt-ansible/issues/82
- [x] Don't run tests if only the docs changed https://github.com/kubevirt/kubevirt-ansible/pull/81

nellyc:
@gbenhaim when do we want to target this? mvp0.1.2 sounds fair?

gbenhaim:
@nellyc  How much time do we have until `mvp0.1.2`?

nellyc:
@gbenhaim approx 2 months

gbenhaim:
We can have it in 2 months.

PR #83, opened 2018-02-20
Use std-ci mirrors

gbenhaim:
- When $STD_CI_YUMREPOS environmet variable is defined, the VMs
  will be configured with the Centos repos that std-ci mirrors.

- RPM installation from the mirrors is faster (the Jenkins slave and the
  mirrors server aer in the same DC.

- It will mitigate errors which occurs when Centos' repos are not
  available.

closes https://github.com/kubevirt/kubevirt-ansible/issues/82
part of https://github.com/kubevirt/kubevirt-ansible/issues/44

### DIFF
--- a/automation/check-patch.openshift_3-7.yumrepos
+++ b/automation/check-patch.openshift_3-7.yumrepos
@@ -1,0 +1,1 @@
+check-patch.yumrepos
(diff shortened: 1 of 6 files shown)


gbenhaim:
Seems like a bug in the service-catalog.
We should use a fixed version for the service catalog, and not latest (which is the default) :-\

PR #84, opened 2018-02-21
re-org: Move openshift deployment to playbooks dir

gbenhaim:
- Custom variables per Openshift version can be declared under
  playbooks/cluster/openshift/vars/

- Use a static version for Openshift 3.9 (v3.9.0-alpha.4)

- Renamed deploy-openshift.yml to config.yml.

- Added the prerequisites of openshift to config.yml - there is no
  reason to keep them in a separate file.


gbenhaim:
If you think that this change is ok, I will also send a PR for updating the docs.

playbooks/cluster/openshift/config.yml
```diff
@@ -2,18 +2,57 @@
...
+        policy: targeted
+        state: enforcing
+
+    - name: stop and disable firewalld
+      register: result
+      service:
+        state: stopped
+        enabled: no
+        name: firewalld
+      failed_when: "result|failed and not 'Could not find the requested service' in result.msg"
```
alexxa: will not `result|failed` be enough?

playbooks/cluster/openshift/config.yml
```diff
@@ -2,18 +2,57 @@
...
       import_role:
         name: "{{ openshift_ansible_dir }}/roles/openshift_facts"
+
     - name: Load openshift facts
       openshift_facts:
         role: common
+
     - name: Set facts for docker-storage-setup
       block:
         - name: Detecting Operating System from ostree_booted
```
alexxa: s/Detecting/Detect

playbooks/cluster/openshift/config.yml
```diff
@@ -2,18 +2,57 @@
...
+      register: result
+      service:
+        state: stopped
+        enabled: no
+        name: firewalld
+      failed_when: "result|failed and not 'Could not find the requested service' in result.msg"
+
+    - name: install epel repository
+      yum:
+        name: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
alexxa: I would use a `epel_repo_url` variable

alexxa: and I'm not sure about the syntax here, see: http://docs.ansible.com/ansible/latest/yum_repository_module.html#id3

playbooks/cluster/openshift/config.yml
```diff
@@ -2,18 +2,57 @@
...
+        - net-tools
+        - bind-utils
+        - iptables-services
+        - bridge-utils
+        - bash-completion
+        - kexec-tools
+        - sos
+        - psacct
+        - docker
+
```
dyasny: I deployed a cluster using the standard CentOS image from https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2

To get openshift-ansible to run, I had to install:
yum-utils ansible wget git net-tools bind-utils iptables-services bridge-utils bash-completion kexec-tools sos psacct docker sshpass NetworkManager centos-release-openshift-origin.noarch ceph-common docker centos-release-openshift iptables-utils NetworkManager atomic-openshift-utils

And ensure NetworkManager is started and enabled

control.yml
```diff
@@ -1,3 +1,3 @@
 - import_playbook: "{{ playbook_dir }}/deploy-with-{{ provider | default('lago') }}.yml"
```
lukas-bednar: I would move all playbooks (including control.yml) under `playbooks` dir

gbenhaim: I will be done in the following patches.

gbenhaim: I didn't want to do the entire re-org in one patch because it makes the code review overwhelming.

playbooks/cluster/openshift/config.yml
gbenhaim: Prerequisites that don't appear here (such as NetworkManager, and centos-release-openshift) are being cared as part of the openshift-ansible modules.

playbooks/cluster/openshift/config.yml
gbenhaim: No, in case that firewalld is not installed, the result|failed is true, but we don't care because firewalld is not installed.

playbooks/cluster/openshift/config.yml
gbenhaim: Done

playbooks/cluster/openshift/config.yml
dyasny: Didn't work for me until I manually installed them before running deploy_cluster.yml

playbooks/cluster/openshift/config.yml
```diff
@@ -2,31 +2,74 @@
...
   pre_tasks:
+    - name: include_vars
+      include_vars:
+        file: "{{ item }}"
+      with_items:
+        - "{{ playbook_dir }}/../../../vars/global_vars.yml"
+        - "{{ playbook_dir }}/vars/{{ openshift_ver }}.yml"
+
+    - name: set SELinux to permissive mode
+      command: setenforce 1
```
rthallisey: This is setting to enforcing mode.  Did you mean ```setenforce 0```? Or is the text wrong?

playbooks/cluster/openshift/config.yml
```diff
@@ -2,31 +2,74 @@
...
+
+    - name: set SELinux to permissive mode under configuration file
```
rthallisey: Can you run this without ```become: true```?

gbenhaim: You can see that in the top of the playbook that we use `remote_user: root`.
If you think that we should use `become`, we can do it a different PR.

playbooks/cluster/openshift/config.yml
gbenhaim: Correct, I'll fix it.

gbenhaim:
@rthallisey I've fixed the `selinux` configuration.

lukas-bednar:
This change is not reflected in README, and made it outdated, please try to keep readme uptodate.
#88

PR #87, opened 2018-02-22
Inventory updates

gbenhaim:
oc: Don't install unneeded features

- Drop manageiq - we don't use manageiq for administrating the
  cluster.
- Don't install examples  - we don't use it.

ASB: Whitelist APBs from docker hub


lukas-bednar:
Maybe guys from miq will need it in future,  @vatsalparekh can you please update us what is plan about miq & kubevirt ?

gbenhaim:
@lukas-bednar If it will be needed in the future, I will add it back.

PR #88, opened 2018-02-23
re-org: Fix readme to reflect re-org changes

lukas-bednar:
Signed-off-by: Lukas Bednar <lbednar@redhat.com.com>


README.md
```diff
@@ -88,7 +88,7 @@ Follow [docker-storage-setup] documentation for more details.
 ```bash
 $ ansible-playbook -i inventory \
     -e "openshift_ansible_dir=openshift-ansible/ \
-    openshift_playbook_path=playbooks/byo/config.yml" deploy-openshift.yml
+    openshift_playbook_path=playbooks/byo/config.yml" playbooks/cluster/openshift/config.yml
```
gbenhaim: With the new playbook, the `openshift_ver` (which indicates which version of openshift do deploy) should be specified. Do you think that we should add a default version?

lukas-bednar: I think it should be documented for sure, I will add this parameter there.

gbenhaim:
Nice !

PR #92, opened 2018-02-25
re-org: mv control.yml

gbenhaim:
Rename control.yml to check-patch.yml and move it into
playbooks/automation.


alexxa:
@gbenhaim It seems like these changes are not reflected in Readme, are they?

gbenhaim:
@alexxa I'll update the README once I'll finish with the re-org.

PR #67, opened 2018-02-14
Playbook to create kubevirt-resources for openshift

rthallisey:
- This PR introduces the new format discussed in https://github.com/kubevirt/kubevirt-ansible/issues/32
 - Using the name ```provision.yaml``` will be compatible with the APB

Deploy kubevirt on an openshift cluster looks like:
   ```ansible-playbook -i localhost playbooks/kubevirt-resources.yml -e @variables/all.yaml```

PR #93, opened 2018-02-26
re-org: mv K8S files

gbenhaim:
- Move K8S roles to the roles dir.
- Move K8S playbook to playbooks/cluster/kubernetes and renamed it to
  config.yml.
- Dropped Kubevirt installation from K8S' playbook, it doesn't belong to
  it.


gbenhaim:
@rthallisey We can use https://github.com/kubevirt/kubevirt-ansible/pull/67 for deploying Kubevirt on K8S.

PR #95, opened 2018-02-26
Fix bad link

rthallisey:


gbenhaim:
@rthallisey Not related to this PR, but do you know if there is a standard for documenting playbooks and roles? I'm talking about docstrings style documentation and not a documentation in an external file.

README.md
```diff
@@ -109,7 +109,8 @@ Currently we don't have a playbook which installs KubeVirt on a Kubernetes clust
 ### OpenShift cluster
 
 Install KubeVirt on your OpenShift cluster.
-[KubeVirt parameters doc](https://github.com/rthallisey/kubevirt-ansible/blob/master/roles/kubevirt/README.md)
+
+[KubeVirt parameters doc](https://github.com/kubevirt/kubevirt-ansible/blob/master/roles/kubevirt/README.md)
```
alexxa: Don't write `doc` and similar in documentation please. It's a slang world.
The explanation of options/parameters/variables goes after a command.
I would replace `KubeVirt parameters doc` with something like `[See](link here) the list of used parameters.`

alexxa:
@gbenhaim [Ansible Proposal: Define a standard way for roles and playbook documentation](https://github.com/ansible/proposals/issues/19)

gbenhaim:
@alexxa too bad that it's just a proposal :-\

README.md
lukas-bednar: Please use relative link when you reference something inside of repository.
Like `[KubeVirt parameters](./roles/kubevirt/README.md)` when repository is moved / renamed / cloned it will always reference proper file.

rthallisey:
@gbenhaim I'm not aware of any standard. I've seen projects do it all sorts of different ways.

README.md
```diff
@@ -109,7 +109,8 @@ Currently we don't have a playbook which installs KubeVirt on a Kubernetes clust
 ### OpenShift cluster
 
 Install KubeVirt on your OpenShift cluster.
-[KubeVirt parameters doc](https://github.com/rthallisey/kubevirt-ansible/blob/master/roles/kubevirt/README.md)
+
+See [KubeVirt parameters documentation](./roles/kubevirt/README.md) for more details on parameters.
```
alexxa: @rthallisey Please move this line after the command which uses these parameters. Command's options/variables explanation should go after a command, not before.

PR #97, opened 2018-02-26
Remove unneeded files

gbenhaim:
- playbooks/components/* are not needed because we changed the way
  kubevirt is deployed.

- playbooks/deploy-cluster.yml and roles/test/t are not being used
  anywhere.


gbenhaim:
@lukas-bednar done

README.md
```diff
@@ -115,8 +115,6 @@ Install KubeVirt on your OpenShift cluster.
 $ ansible-playbook -i localhost playbooks/kubevirt.yml -e @variables/all.yaml
 ```
 
-[![asciicast](https://asciinema.org/a/161278.png)](https://asciinema.org/a/161278)
```
lukas-bednar: I think we should remake demo ... we can start with opening issue for that. What do you think ?

gbenhaim: I think that we can start with opening an issue. 
But let's wait until we finish the re-org before recording it.

PR #98, opened 2018-02-27
Drop variables/all.yaml

gbenhaim:
It's not needed since we have the default vars of kubevirt's role.


rthallisey:
@gbenhaim @lukas-bednar 
I was a little late to review this, but IMO we should keep a top level variable file for folks to override variables.  This will come in handy when we only want to expose a subset of variables.  For example when we add components, we will want to expose a variable like ```enable_cinder: true``` at the top level.  But, at a lower level, something like ```dev_template_resources:```, which is not likely to change, we leave in ```defaults/main.yml```

Issue #101, opened 2018-02-28
Failure to log in to the server on playbooks/kubevirt.yml


alexxa:
playbooks/kubevirt.yml returns login failure.

```  

>> ansible-playbook -i localhost playbooks/kubevirt.yml
<...>
TASK [kubevirt : Create kube-system namespace] *********************************************************
Wednesday 28 February 2018  10:09:05 +0100 (0:00:00.345)       0:00:00.453 **** 
 [WARNING]: when statements should not include jinja2 templating delimiters such as {{ }} or {% %}.
Found: ns.stdout != "{{ namespace }}"

fatal: [localhost]: FAILED! => {
    "changed": true,
    "cmd": "kubectl create namespace kube-system",
    "delta": "0:00:00.097699",
    "end": "2018-02-28 10:09:05.327740",
    "rc": 1,
    "start": "2018-02-28 10:09:05.230041"
}

STDERR:

error: You must be logged in to the server (Unauthorized)

```

gbenhaim:
You should login to the cluster by yourself or specify the following variables:

```
admin_user (a user with cluster-admin permissions)                                                                                                    
admin_password
```

alexxa:
@gbenhaim There is nth like that in docs, in our Readme, and if it's that way, then our docs should be very clear on that, agree?

gbenhaim:
@rthallisey Should it be added to the docs of kubvirt's role?

alexxa:
I think, it should be right before a command which triggers playbooks/kubevirt.yml
```
Before running KubeViert deployment, ensure you are logged in your OpenShift cluster as a user with cluster-admin permissions.

$ oc login <master_fqdn>:<port>
Username:
Password:


```

BTW, using ansible-playbook and providing user/pwd as extra variables disclosures cluster login credentials. One can see them if they list process (`ps -aux | grep ansible-playbook`) or get access to `history`. What do you think about it?

rthallisey:
If you run ```ansible-playbook -i localhost playbooks/kubevirt.yml``` the expectation is you are logged into the cluster.  We'll need to mention this is the docs.

alexxa:
@gbenhaim @rthallisey @lukas-bednar And what do you think about login credentials disclosure I mentioned in my previous comment?

Issue #100, opened 2018-02-28
kubectl dependency on KubeVirt installation on an existing cluster 


alexxa:
After successful installation of OpenShift cluster (3.7), following [Readme](https://github.com/kubevirt/kubevirt-ansible#openshift-cluster-1), I run into

```
>> ansible-playbook -i localhost playbooks/kubevirt.yml
<...>

TASK [kubevirt : Create kube-system namespace] *********************************************************
Wednesday 28 February 2018  09:57:19 +0100 (0:00:00.254)       0:00:00.361 **** 
 [WARNING]: when statements should not include jinja2 templating delimiters such as {{ }} or {% %}.
Found: ns.stdout != "{{ namespace }}"

fatal: [localhost]: FAILED! => {
    "changed": true,
    "cmd": "kubectl create namespace kube-system",
    "delta": "0:00:00.001448",
    "end": "2018-02-28 09:57:19.575954",
    "rc": 127,
    "start": "2018-02-28 09:57:19.574506"
}

STDERR:

/bin/sh: kubectl: command not found


MSG:

non-zero return code

```

Should I have `kubectl` installed? If so, it should be documented. But the best is to include it in the playbook. After installing `origin-clients` by `dnf` the playbooks proceeds normally.

gbenhaim:
Installing it a part of the playbook will require running with sudo.
We can add a test the check if `kubectl` exists, if not fail the playbook with an informative message.

alexxa:
Whatever way it will be solved, a `kubectl` requirement note in docs should be there and will be appreciated.

lukas-bednar:
@alexxa could you please cover this gap in documentation ?

rthallisey:
> Whatever way it will be solved, a kubectl requirement note in docs should be there and will be appreciated.
>
Definitely should be added to the docs.

alexxa:
@lukas-bednar That was a plan, but @rthallisey beat me.

PR #104, opened 2018-02-28
Document logging into an OpenShift cluster and location of kubectl config

rthallisey:
Fixes: #101 
Fixes: #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ KubeVirt Ansible consists of a set of Ansible playbooks that deploy fully functi
 
 To deploy KubeVirt on an existing OpenShift cluster run the command below. For more information on clusters and other deployment scenarious see [playbooks instructions](./playbooks/README.md).
 
+
 ```
 ansible-playbook -i localhost playbooks/kubevirt.yml -e@vars/all.yml
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -102,6 +102,9 @@ where
 
 ## Install KubeVirt on an existing cluster
 
+Make sure you have a kubeconfig file that's available for ```kubectl``` to use.
```
lukas-bednar: Issue 100 was about missing kubectl command, I would mention in addition that `kubectl` has to be ensured.

rthallisey:
@alexxa any more comments here?

alexxa:
@rthallisey Yes, I mentioned a comment of @gbenhaim before: "We can add a test the check if kubectl exists, if not fail the playbook with an informative message." I think it can be a nice enhancement in addition to a doc note. Will you agree?

rthallisey:
@alexxa I think a doc should be enough because the error message from #100  ```/bin/sh: kubectl: command not found``` is understandable that you don't have kubectl installed.

alexxa:
@rthallisey Please update this PR, and move all the changes to playbooks/README.md

playbooks/README.md
```diff
@@ -97,6 +96,10 @@ Currently we don't have a playbook which installs KubeVirt on a Kubernetes clust
 
 ### OpenShift cluster
 
+The playbook expects you have the client installed and a config file in
```
alexxa: @rthallisey what client? and a link on installation step docs will be appreciated.

alexxa:
Thank you! LGTM. Merging.

Issue #115, opened 2018-03-01
Use a top level variables file


rthallisey:
The goal of a top level variables file is to provide the user with a single file that lists the top level variables of kubevirt-ansible.  The variables are mostly playbook level variables that will have a significant impact on how playbooks will run.  Some variables listed here can also be variables that users are likely to change or variables we want a user to be aware of.

```yaml
### Cluster ###
cluster: openshift
namespace: kube-system

### KubeVirt ###
manifest_version: release
docker_tag: latest

### Storage ###
enable_cinder_ceph: false
enable_gluster: false
```

PR #116, opened 2018-03-01
Add a top level vars file and add some defaults to roles

rthallisey:
Fixes: #115


playbooks/cluster/openshift/README.md
```diff
@@ -0,0 +1,10 @@
+# OpenShift Cluster Role
```
lukas-bednar: s/Role/Playbook/ it seems to be playbook to me :-)

playbooks/provider/lago/README.md
```diff
@@ -0,0 +1,13 @@
+# Logo Provider Role
```
lukas-bednar: s/Logo/Lago/, and s/Role/Playbook/

automation/check-patch.sh
```diff
@@ -145,9 +145,9 @@ main() {
         "mode=$mode"
         "provider=$provider"
         "inventory_file=$inventory_file"
-        "cluster_type=$cluster_type"
+        "cluster=$cluster"
         "ansible_modules_version=$ansible_modules_version"
-        "openshift_ver=$openshift_ver"
+        "openshift_version=$openshift_version"
```
gbenhaim: `openshift_version` is also being used in `openshift-ansible`. I think that we shouldn't use it also.

playbooks/provider/lago/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+cluster: "openshift"
```
gbenhaim: Except for `inventory`, all the other variables are not being used by the Lago provider.
Why do we need this file?

vars/all.yml
```diff
@@ -0,0 +1,12 @@
+### Cluster ###
```
gbenhaim: I agree that this file is needed, but:

- It should be near `kubevirt.yml`, which is the playbook that will use it.
- We should include this file using `include_vars` from `kubevirt.yml`.
- This file shouldn't contain any variables that are not related to Kubevirt's deployment. You can place global 
  variables in `vars/global_vars.yml`, and I think that we shouldn't use this file often.

rthallisey: The goal is for this file to be the top level interface for kubevirt-ansible, not just the kubevirt playbook.  Variables declared here like openshift_version and cluster will be used in multiple playbooks.  This should absorb ```vars/global_vars.yml```.

We can use ```include_vars``` in the playbook so we don't have to use ```-e@vars/all.yml```.

vars/all.yml
```diff
@@ -0,0 +1,12 @@
...
+cluster: openshift
+namespace: kube-system
+
+# OpenShift #
+openshift_version
```
rthallisey: /me need to add a value for openshift_version

automation/check-patch.sh
rthallisey: Let's keep it so we expose it to the top level

rthallisey:
@lukas-bednar @gbenhaim any more comments? Would like to get this merged asap

playbooks/cluster/openshift/config.yml
```diff
@@ -1,13 +1,15 @@
 ---
 - hosts: all
   remote_user: root
+  vars:
+    epel_release_rpm_url: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
alexxa: @rthallisey I'm not an Ansible expert, and I don't follow why it's here if there is the same var in playbooks/cluster/openshift/defaults/main.yaml ?

rthallisey: Something I noticed in CI was that the variables in defaults are not being used.  But when I run the playbook locally they do.  So to answer your first question, I tried changing the name from yml to yaml, but it didn't work.  Finally I added the variable here for now to see if CI will go back to green and it did.

I'll investigate this some more tomorrow.

PR #120, opened 2018-03-05
[WIP] storage: Add a storage-demo role

aglitke:
The storage-demo role deploys a self-contained ephemeral storage
environment consisting of a cider and ceph.  This should only be used
by developers.  All data stored will be lost when the storage-demo
statefulset is terminated.

### DIFF
--- a/playbooks/automation/check-patch.yml
+++ b/playbooks/automation/check-patch.yml
@@ -4,3 +4,4 @@
 - import_playbook: "{{ playbook_dir }}/../cluster/{{ cluster | default('openshift') }}/config.yml"
 - import_playbook: "{{ playbook_dir }}/../cluster-login.yml"
 - import_playbook: "{{ playbook_dir }}/../kubevirt.yml"
+- import_playbook: "{{ playbook_dir }}/../storage.yml"
--- a/playbooks/storage.yml
+++ b/playbooks/storage.yml
@@ -1,0 +1,22 @@
+---
+- hosts: localhost
+  vars:
+    storage_role: storage-demo
+  connection: local
+  gather_facts: False
+  # unset http_proxy. required for running in the CI
+  environment:
+    http_proxy: ""
+  roles:
+    - role: "{{ storage_role }}"
+
+- hosts: masters nodes
+  vars:
+    storage_role: storage-demo
+  gather_facts: False
+  # unset http_proxy. required for running in the CI
+  environment:
+    http_proxy: ""
+  roles:
+    - role: storage-demo-nodeconfig
+      when: storage_role == 'storage-demo'
--- a/roles/storage-demo-nodeconfig/README.md
+++ b/roles/storage-demo-nodeconfig/README.md
@@ -1,0 +1,9 @@
+# Ansible role: storage-demo-nodeconfig
+
+**Note: This role is not for production use.  All storage is erased any time**
+**the storage-demo pod is stopped.**
+
+This role configures the nodes in a cluster to work with the storage-demo role.
+Currently this means adjusting the firewall to allow ceph traffic to flow
+between nodes.  As such, this will only work on a cluster that is running
+iptables.
--- a/roles/storage-demo-nodeconfig/tasks/deprovision.yml
+++ b/roles/storage-demo-nodeconfig/tasks/deprovision.yml
@@ -1,0 +1,19 @@
+---
+- name: "Remove rule: Allow ceph OSD traffic"
+  iptables:
+    state: absent
+    chain: INPUT
+    protocol: tcp
+    destination_port: 6789
+    jump: ACCEPT
+
+- name: "Remove rule: Allow ceph MDS traffic"
+  iptables:
+    state: absent
+    chain: INPUT
+    protocol: tcp
+    destination_port: 6800:7300
+    jump: ACCEPT
+
+- name: Save iptables configuration
+  command: iptables-save
--- a/roles/storage-demo/README.md
+++ b/roles/storage-demo/README.md
@@ -1,0 +1,21 @@
+# Ansible role: storage-demo
+
+**Note: This role is not for production use.  All storage is erased any time**
+**the storage-demo pod is stopped.**
+
+This role deploys a self-contained storage environment suitable for development
+and testing.  A single-instance ephemeral ceph cluster is created inside a pod.
+Cinder is deployed and connected to ceph.  A dynamic provisioner and related
+kubernetes resources are created to interface with the cluster.
+
+## Configuration parameters
+* **namespace**: The namespace into which the storage-demo components should be
+  installed
+* **cluster**: The type of cluster (openshift or kubernetes)
+* **cinder_provisioner_repo**: The repository containing the cinder provisioner
+* **cinder_provisioner_release**: The docker image tag to use for the cinder
+  provisioner
+* **action**: The action to perform.  Currently only 'provision' is supported
+* **storage_demo_template_dir**: The location of the deployment template file.
+  There is no need to change this value.
+  
--- a/roles/storage-demo/templates/storage-demo.yml
+++ b/roles/storage-demo/templates/storage-demo.yml
@@ -1,0 +1,128 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: storage-demo-cluster-admin
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  name: cluster-admin
+  kind: ClusterRole
+subjects:
+- kind: ServiceAccount
+  name: storage-demo
+  namespace: {{ namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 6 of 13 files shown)


roles/storage-demo/defaults/main.yaml
```diff
@@ -0,0 +1,8 @@
+namespace: "kube-system"
+cluster: "openshift"
+cinder_provisioner_repo: "quay.io/aglitke"
+cinder_provisioner_release: "latest"
+storage_demo_release: "latest"
+# The action to perform: provision
+action: "provision"
+storage_demo_template_dir: "{{ role_path }}/templates"
```
rthallisey: Can you add all these variables to a README in  ```roles/storage-demo/```

roles/storage-demo/defaults/main.yaml
```diff
@@ -0,0 +1,8 @@
+namespace: "kube-system"
+cluster: "openshift"
+cinder_provisioner_repo: "quay.io/aglitke"
+cinder_provisioner_release: "latest"
+storage_demo_release: "latest"
+# The action to perform: provision
```
alexxa: Is there a reason why this particular variable is commented and others not?

rthallisey:
To add this to automation, you'll need to add your playbook to https://github.com/kubevirt/kubevirt-ansible/blob/master/playbooks/automation/check-patch.yml.

I also think we want to merge https://github.com/kubevirt/kubevirt-ansible/pull/116 first, so we can add component variables to ```vars/all.yml```.

playbooks/storage-demo.yaml
```diff
@@ -0,0 +1,8 @@
+- hosts: localhost
+  connection: local
+  gather_facts: False
+  # unset http_proxy. required for running in the CI
+  environment:
+    http_proxy: ""
+  roles:
```
rthallisey: Let's make this configurable.
```
  vars:
      storage_role: demo
  roles:
      - role: {{ storage_role }}
```

rthallisey: Also rename to ```storage.yml```

roles/storage-demo/templates/storage-demo.yml
```diff
@@ -0,0 +1,126 @@
...
+metadata:
+  name: storage-demo-cluster-admin
+roleRef:
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: storage-demo
+  namespace: {{ namespace }}
+---
+apiVersion: authorization.openshift.io/v1
```
rthallisey: Use ```rbac.authorization.k8s.io/v1``` so this will work for both openshift & k8s

roles/storage-demo/templates/storage-demo.yml
```diff
@@ -0,0 +1,126 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: storage-demo-default-admin
+roleRef:
+  name: cluster-admin
```
rthallisey: Change this to:
```
roleRef:
   kind: ClusterRole
   name: cluster-admin
   apiGroup: rbac.authorization.k8s.io
```

aglitke: Thanks!

playbooks/automation/check-patch.yml
```diff
@@ -4,3 +4,4 @@
 - import_playbook: "{{ playbook_dir }}/../cluster/{{ cluster | default('openshift') }}/config.yml"
 - import_playbook: "{{ playbook_dir }}/../cluster-login.yml"
 - import_playbook: "{{ playbook_dir }}/../kubevirt.yml"
+- import_playbook: "{{ playbook_dir }}/../storage.yml"
```
gbenhaim: I think that we should add a new playbook that deploys the different components, and import it in check-path.yml. It will allow users to deploy all the different components on an existing cluster.

roles/storage-demo-nodeconfig/tasks/deprovision.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: "Remove rule: Allow ceph OSD traffic"
```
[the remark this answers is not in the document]
gbenhaim: @rthallisey How does firewall configuration is being cared when deploying using an APB?

playbooks/storage.yml
```diff
@@ -0,0 +1,22 @@
...
+- hosts: localhost
+  vars:
+    storage_role: storage-demo
+  connection: local
+  gather_facts: False
+  # unset http_proxy. required for running in the CI
+  environment:
+    http_proxy: ""
+  roles:
+    - role: "{{ storage_role }}"
```
gbenhaim: Shouldn't we first call `storage-demo-nodeconfig`?

roles/storage-demo/templates/storage-demo.yml
```diff
@@ -0,0 +1,128 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  name: cluster-admin
+  kind: ClusterRole
+subjects:
+- kind: ServiceAccount
+  name: storage-demo
+  namespace: default
+---
+apiVersion: apps/v1
```
rthallisey: ```apiVersion: v1```

roles/storage-demo-nodeconfig/tasks/deprovision.yml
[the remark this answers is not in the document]
rthallisey: This is split from the other role because it runs tasks that change the hosts.  For security reasons, APBs don't change anything on the hosts.  Changing this is on our radar though, but it needs to come after the service-catalog adds rbac for apbs so we can give apbs that have cluster-admin this level of host and cluster access.

roles/storage-demo/templates/storage-demo.yml
rthallisey: Scratch that, looks like we need ```apiVersion: v1``` ```when openshift_version==3.9```

playbooks/automation/check-patch.yml
rthallisey: We're you thinking a components playbooks?  I think we can look at doing that in another PR.  I think there are multiple good approaches here.

roles/storage-demo-nodeconfig/README.md
```diff
@@ -0,0 +1,8 @@
+# Ansible role: storage-demo-nodeconfig
+
+**Note: This role is not for production use.  All storage is erased any time**
+**the storage-demo statefulset is stopped.**
+
+This role configures the nodes in a cluster to work with the storage-demo role.
+Currently this means adjusting the firewall to allow ceph traffic to flow
```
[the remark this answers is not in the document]
aglitke: ok.  Updated README.md

roles/storage-demo-nodeconfig/tasks/deprovision.yml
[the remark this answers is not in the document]
aglitke: I am trying to use a formulaic method for all storage roles based on what is being done for the kubevirt role.  Therefore I followed the pattern of using the action parameter to select the playbook to run.  The storage-demo and storage-demo-nodeconfig roles need to be separate because they target separate hosts.  The former runs on localhost and fits nicely into an APB.  The later runs on the nodes themselves and may have issues as an APB.  We might move the nodeconfig into the openshift deployment flow and include the role only when the storage_role == 'storage-demo'.

playbooks/automation/check-patch.yml
aglitke: can't this be done today by simply running the desired playbooks in order?  I guess this is a general comment to make the repo easier for everyone to use and can be handled separately?

playbooks/storage.yml
aglitke: The order isn't important.  Both roles need to be applied before the storage can actually be used.

aglitke:
@alexxa and @gbenhaim I have made some of your suggested changes and commented on others.  Are you OK with merging this PR at this point?

roles/storage-demo-nodeconfig/README.md
```diff
@@ -0,0 +1,9 @@
+# Ansible role: storage-demo-nodeconfig
+
+**Note: This role is not for production use.  All storage is erased any time**
```
alexxa: s/any time/every time

roles/storage-demo/README.md
```diff
@@ -0,0 +1,21 @@
+# Ansible role: storage-demo
+
+**Note: This role is not for production use.  All storage is erased any time**
```
alexxa: s/any time/every time

roles/storage-demo-nodeconfig/README.md
aglitke: I think these are interchangeable in this context so can we keep it if this is the only requested change?  I can reword if I respin the patch for another reason.

roles/storage-demo/README.md
aglitke: I think these are interchangeable in this context so can we keep it if this is the only requested change?  I can reword if I respin the patch for another reason.

rthallisey:
I'm good with merging this as.  We can always come back and shift things around if folks feel we need to. Nice work @aglitke.

Issue #128, opened 2018-03-08
Dynamic inventory


aglitke:
As part of the adding a cns storage flavor, I would like to add some things to the inventory file to conditionally deploy cns.  If cluster == openshift and storage_role == storage_cns I'd want to make the following changes to the inventory.  What is the best way to achieve this?  

```
[OSEv3:children]
...
glusterfs
```
```
[OSEv3:vars]
...
# Namespace for CNS pods (will be created)
openshift_storage_glusterfs_namespace=app-storage
# Automatically create a StorageClass referencing this CNS cluster
openshift_storage_glusterfs_storageclass=true
# glusterblock functionality is not supported outside of Logging/Metrics
openshift_storage_glusterfs_block_deploy=false
# Disable any other default StorageClass
openshift_storageclass_default=false
```

```
[glusterfs]
<master> glusterfs_devices='[ "/dev/vdd" ]'
<node0> glusterfs_devices='[ "/dev/vdd" ]'
<node1> glusterfs_devices='[ "/dev/vdd" ]'
...
```

lukas-bednar:
I am not sure if I understand why we need dynamic inventory.

I think you can add these things into `./inventory` file right a way.
And we could have variable there which controls whether to deploy cns or not.

Regarding to automation, lago generates inventory file containing hosts groups, and in order to add our VMs into `glusterfs` host group, it is enough to change these two lines in LagoInitFile.yml:

For nodes:
https://github.com/kubevirt/kubevirt-ansible/blob/08cca1baacbc164499c0882522ce3676c8a9e725/playbooks/provider/lago/LagoInitFile.yml#L16
And for master
https://github.com/kubevirt/kubevirt-ansible/blob/08cca1baacbc164499c0882522ce3676c8a9e725/playbooks/provider/lago/LagoInitFile.yml#L41

aglitke:
@lukas-bednar We will still need to generate the glusterfs_devices=[ "/dev/vdd" ] variable for each host in the glusterfs group.  I don't believe your code change would achieve that part?

@rthallisey We can decide which is the default but I think it would be best to only run the gluster provisioning steps in openshift-ansible when we intend to use them.  Whether those steps are run is dependent entirely on the contents of the inventory file.

lukas-bednar:
@aglitke hm, I understand that name of device can be different on each machine in real life.
[...]

PR #129, opened 2018-03-08
Add storage-cns role

aglitke:
Add a storage-cns role to connect kubevirt with existing CNS storage in
the cluster.  This role does not install CNS on a cluster.  It assumes
that CNS has been installed via one of the already supported
installation methods.

### DIFF
--- a/roles/storage-cns/README.md
+++ b/roles/storage-cns/README.md
@@ -1,0 +1,13 @@
+# Ansible role: storage-cns
+
+This role deploys cluster resources necessary for kubevirt to interface with
+CNS (glusterfs) storage.  This role assumes that CNS itself has already been
+installed.
+
+### Variables
+| Variable        | Default Value           | Description  |
+|:------------- |:-------------|:----- |
+| admin_user | _optional_ | User with cluster-admin permissions. Used in the APB |
+| admin_password | _optional_ | Password for cluster-admin user. Used in the APB |
+| cluster | openshift | The cluster we're running on |
+| action | provision | The action of provisioning or deprovisioning CNS add-ons |
--- a/roles/storage-cns/tasks/deprovision.yml
+++ b/roles/storage-cns/tasks/deprovision.yml
@@ -1,0 +1,14 @@
+---
+- name: Login As Super User
+  command: "oc login -u {{ admin_user }} -p {{ admin_password }}"
+  when: cluster=="openshift"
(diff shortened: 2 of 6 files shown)


aglitke:
This role will not be tested by CI.  In order to do that we must decide whether we enable permutations of the storage_role variable and whether to update the inventory file as being discussed in #128 .

roles/storage-cns/README.md
```diff
@@ -0,0 +1,11 @@
+# Ansible role: storage-cns
+
+This role deploys cluster resources necessary for kubevirt to interface with
+CNS (glusterfs) storage.  This role assumes that CNS itself has already been
+installed.
+
+## Configuration parameters
```
rthallisey: nit: Can you make this a table like the other READMEs

roles/storage-cns/tasks/deprovision.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Login As Super User
+  command: "oc login -u {{ admin_user }} -p {{ admin_password }}"
+  when: cluster=="openshift"
+        and admin_user is defined
+        and admin_password is defined
+
+- name: Render storage-cns deployment yaml
+  template:
+    src: "{{ storage_cns_template_dir }}/storage-cns.yml"
```
rthallisey: Above it says ```There is no need to change this value```. By default, template will look in the ```templates``` directory so you drop this variable and change to ```src: storage-cns.yml```.

roles/storage-cns/README.md
aglitke: Sure.

roles/storage-cns/tasks/deprovision.yml
aglitke: Thanks.  Fixed.

lukas-bednar:
Looks good to me!
I would be happy if it could be part of CI process, and work with #130 .
According to jenkins gluster was deployed successfully.
http://jenkins.ovirt.org/blue/rest/organizations/jenkins/pipelines/kubevirt_kubevirt-ansible_standard-check-pr/runs/254/nodes/92/steps/163/log/?start=0

aglitke:
@lukas-bednar I agree we want to wire it up.  Can we merge this now (so it can be used in an APB) and wire it up once #130 is merged?

PR #139, opened 2018-03-12
Hookup storage playbook to kubevirt playbook

rthallisey:
By default we'll disable storage.


vars/all.yml
```diff
@@ -10,3 +10,4 @@ manifest_version: release
 docker_tag: latest
 
 ### Components ###
+enable_storage: no
```
alexxa: Shoud [Readme](https://github.com/kubevirt/kubevirt-ansible/blob/master/roles/kubevirt/README.md) mention that var and it defaults to no?

vars/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Variables
+| Variable        | Default Value           | Description  |
+|:------------- |:-------------|:----- |
+| cluster | openshift | The cluster we're running on |
+| namespace | kube-system | Namespace to create resources |
+| openshift_version | 3.7 | OpenShift cluster version. Either 3.7 or 3.9 |
+| manifest_version | release | KubeVirt manifest version |
+| docker_tag | latest | Container image tag |
+| storage_role | None | Storage role to install with KubeVirt |
```
lukas-bednar: I am little bit confused by name of this variable, I am getting feeling that I should provide name of `storage_role` to be used, and according to playbook I see that `storage.yml` playbook is loaded for any value != None.
I would either improve doc for this variable, or change name.

vars/README.md
```diff
@@ -10,4 +10,16 @@ List of top level variables.
...
+### Storage Roles
+
+**storage-demo**
+All-in-one ephemeral storage.
+
+**CNS**
+```openshift-ansible``` will provide Gluster storage and the CNS role will
+create the storage class.
+
+**Cinder**
```
alexxa: Honestly, I don't like how it looks -how some variables go under the table, what if others appear?  And I don't like hardcodng default values in docs too. ATM I think about two solutions.
1) Can it all be in one table? with Defaults and Possible values?
2) If not, what about writing all possible values in a table in Readme and for default values add a reference to var file?

rthallisey: That's fair.  I think storage-roles could be their own document frankly, but I can list the defaults in the tbale

rthallisey:
ci test please

lukas-bednar:
@rthallisey is there anything missing ? If it is complete I would get it in.

Issue #52, opened 2018-02-02
Why it need to remove firewalld and NetworkManager packages


gouyang:
I'm trying to preparing kubernetes cluster and deploy KubeVirt on my fedora 27 workstation, my system crashed because the playbook removes NetworkManager and all its dependencies. I know it's better to use a server to run the playbooks, but it would be better to be able run it on user's workstation, especially for those who just come to the project.

fabiand:
Did you find out why NM was removed?

gouyang:
The openshift-ansible and kubespray project can also deploy kubernetes cluster without remove these packages. By looking into the the script `kubernetes/roles/prerequisites/tasks/main.yml`, it maybe used for enable the bridge network, but I'm not sure of it.

gouyang:
Remove NetworkManager will remove lots of dependency packages on user's workstation and then the system crash, even not recoverable.

dyasny:
NetworkManager is required by openshift-ansible, has to be installed, enabled and started (seen on 3.7 and 3.9 so far)

gouyang:
> Okay. So if ths happens in the flow wher ekubernetes or openshift are getting deployed, then this is not a bug, because for these deployments it might be necessary.

I was meaning `openshift-ansible` and `kubespray` could deploy kubernetes cluster successfully without remove NM package.

dyasny:
> So it's not necessary to remove NM to deploy kubernetes cluster.

I have tried to deploy openshift-ansible on a set of nodes where NetworkManager was not installed, and it failed with a dependency error. I had to install, start and enable NetworkManager for openshift-ansible to run a successful deployment

gouyang:
`ansible-playbook -i inventory playbooks/cluster/kubernetes/config.yml` need to remove NM as the issue reported.

` ansible-playbook -i inventory -e "openshift_ansible_dir=/root/openshift-ansible openshift_playbook_path=playbooks/byo/config.yml  openshift_ver=3.7" playbooks/cluster/openshift/config.yml` meet below error which need NM installed and enabled.

```
TASK [openshift_node_dnsmasq : fail] *************************************************************************************************************
Thursday 01 March 2018  10:35:08 +0000 (0:00:00.029)       0:05:03.415 ******** 
fatal: [bootp-73-3-160.eng.pek2.redhat.com]: FAILED! => {
    "changed": false
}

MSG:

Currently, NetworkManager must be installed and enabled prior to installation.
```

[...]

PR #143, opened 2018-03-14
Keep NetworkManager installed :bug:

alexxa:
Don't remove NM as issue #52 says.


alexxa:
It also should check that NM is installed and running.

alexxa:
This is ready for review.

PR #151, opened 2018-03-19
Temporary playbook to disable selinux

rthallisey:
Currently SELinux permissive mode is required in order to run this early version of KubeVirt.

We will switch to enforcing mode once KubeVirt is more stable.

### DIFF
--- a/playbooks/README.md
+++ b/playbooks/README.md
@@ -109,6 +109,11 @@ indentified with the cluster using ```oc login```.
 
 Install KubeVirt on your OpenShift cluster.
 
+Before installing KubeVirt on an existing OpenShift cluster, ensure that SELinux is disabled on all hosts:
+```bash
+$ ansible-playbook -i inventory playbooks/selinux.yml
+```
+
 ```bash
 $ ansible-playbook -i localhost playbooks/kubevirt.yml -e@vars/all.yml
 ```
--- a/playbooks/selinux.yml
+++ b/playbooks/selinux.yml
@@ -1,0 +1,8 @@
+- hosts: all
+  gather_facts: no
+  vars:
+    selinux: "permissive"
+  tasks:
+    - name: "setenforce {{ selinux }}"
+      shell: "setenforce {{ selinux }}"
+      become: true
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ KubeVirt Ansible consists of a set of Ansible playbooks that deploy fully functi
 
 ## Contents
 
-* `atomation/`: CI scripts to verify the functionality of playbooks.
+* `automation/`: CI scripts to verify the functionality of playbooks.
 * `playbooks/`: Ansible playbooks to provision resources, deploy a cluster and install KubeVirt for various scenarios.
 * `roles/`: Roles to use in playbooks.
 * `vars/`: Variables to use in playbooks.


rthallisey:
This playbook is temporary.  I think we should keep this separate so we can easily remove it later on.

playbooks/selinux.yml
```diff
@@ -0,0 +1,8 @@
+- hosts: all
+  gather_facts: no
+  vars:
+    selinux: 0
```
fabiand: Please rename this `enforcing` the it is inline with the bootloader arg :)

playbooks/README.md
```diff
@@ -121,6 +121,13 @@ $ ansible-playbook -i inventory -e@vars/all.yml playbooks/automation/check-patch
 See [Lago parameters documentation](./playbooks/provider/lago/README.md) for more details.
 
 
+### Disable SELinux Playbook
+Disable SELinux on all hosts (Temporary playbool).
```
gbenhaim: s/playbool/playbook

rthallisey:
@gbenhaim After you setup a cluster with openshift-ansible

playbooks/README.md
```diff
@@ -132,6 +132,13 @@ create the storage class.
 **Cinder**
 
 
+### Disable SELinux Playbook
+Disable SELinux on all hosts (Temporary playbook).
```
alexxa: It's not clear when it should be executed.

alexxa: And I think it can added simply as a Note:

```
>  **Note:** Text here. E.g.: Before installing KubeVirt on an existing OpenShift cluster, ensure that SELinux is disabled on all hosts: code here
```

gbenhaim:
@rthallisey What is wrong with disabling it in `kubevirt-ansible/playbooks/cluster/openshift/config.yml`?

playbooks/README.md
rthallisey: Agreed that this could be handled by docs, but folks want to have a playbook they can run when setting up a demo environment.  This playbook will temporarily live here.

rthallisey:
@gbenhaim this playbook is only used when folks use another method besides kubevirt-ansible to launch openshift/kubernetes.  The expected workflow is: deploy openshift with openshift-ansible, run disable-selinux playbook, launch kubevirt with the apb

playbooks/README.md
alexxa: @rthallisey I didn't mean replace playbook by doc. I meant update doc to document the usage of that temporary playbook. The current doc string in this PR is not sufficient.  =)) Please reread ^^

alexxa: Line: 
```
  >  **Note:** Before installing KubeVirt on an existing OpenShift cluster, ensure that SELinux is disabled on all hosts:
$ ansible-playbook -i inventory playbooks/selinux.yml

```
should be placed before line: 

https://github.com/rthallisey/kubevirt-ansible/blame/4e8b81dd80e608a76668228bbb7b3de99664d4fe/playbooks/README.md#L103

If I understand it correctly

alexxa: sorry, if I wasn't clear.

rthallisey: will update, thanks @alexxa

nellyc:
@rthallisey I wonder if we can remove this workaround now that https://github.com/kubevirt/kubevirt/issues/639 was resolved? (maybe we need to switch to kubevirt-0.4.0 first)

@lukas-bednar I know we have this w/a in other places, can we remove them as well?

lukas-bednar:
@nellyc Let's see what will happen #180 .
Unfortunately we don't have any test which spawns VM after deployment, so I will have to test it manually anyway.

PR #81, opened 2018-02-20
automation: Run the CI only when the code changed

gbenhaim:
With this change, the CI will run only if a file which
doesn't end with ".md" has been changed.

closes https://github.com/kubevirt/kubevirt-ansible/issues/76

Signed-off-by: gbenhaim <galbh2@gmail.com>

Issue #76, opened 2018-02-19
CI tests should ignore documentation changes


alexxa:
CI should ignore commits containing [ci skip] and [skip ci] in their title or description. [ci skip] and [skip ci] can be used for documentation and minor changes which don't require CI check.

eedri:
We should be very careful with such kind of tags, its a big opening for allowing PRs to get merged w/o CI tests. 

A better solution is for the CI system to detect its a text change and not run validation on it.

gbenhaim:
I agree with @eedri. CI should be triggered for each PR.
In the automation scripts, we can implement a logic which skips the cluster deployment and tests if the code didn't change.

alexxa:
https://github.com/kubevirt/kubevirt-ansible/pull/74#issuecomment-366701197

eedri:
+1 for the issue, we should update the automation/check-patch.sh script to detect documentation.
Another option is to add another sub-stage for testing documentation ( like spell-checker )

PR #153, opened 2018-03-19
WIP: test skipping

eedri:
Signed-off-by: Eyal Edri <eedri@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # KubeVirt Ansible
 
+
+
 __Tools to provision resources, deploy clusters and install KubeVirt.__
 
 ## Overview
--- a/automation/check-patch.sh
+++ b/automation/check-patch.sh
@@ -13,6 +13,7 @@ get_run_path() {
         mkdir -p "/dev/shm/ost" && \
         echo "/dev/shm/ost/deployment-$suffix" || \
         echo "$PWD/deployment-$suffix"
+        echo "End of get_run_path..."
 }
 
 collect_logs() {


eedri:
looks like adding [skip ci] works and doesn't trigger CI.

eedri:
dropping the [skip ci] from title to check

eedri:
ci test please

alexxa:
@eedri CI wasn't triggered because you updated md doc. CI ignores changes in docs, see PR #81. Update code file, use [skip ci] and try again.

eedri:
I don't think this is correct, the feature to not run CI if you update .md files is in the check-patch.sh script, which means Jenkins job has to run to get to that code.
In this case no job run, so its not possible it could know which file was changed.

But I'll update code as well to prove it.

alexxa:
This PR was updated 3 times (doc change skip ci, title change, code change skip ci). I see 2 jenkins jobs were started.

eedri:
They were started because I removed the skip CI on previous PR, did it
trigger again on the last one as well ?

alexxa:
I do think so. Because I remember that I pushed my first commit on this repo, and used [skip ci], but CI was triggered anyway. Then I tried [ci skip], it din't work as I expected too. I opened an issue about it: #76 and originally titled it "add [ci skip] and [skip ci] in CI" and see it's description - meaning [skip ci] flag didn't work for me back then. Does it really work for you now? If so, I don't understand how and what changed since I reported that issue.

eedri:
Hi,
I can think on a few options why it didn't pass in the past.
First, is it possible that [skip-ci] was used? i.e with a dash(-)? because according to the regex, it looks like only whitespace is supported.

Second, we recently upgraded the GitHub plugin on Jenkins, and I remeber it did have issues, so maybe the upgrade fixed it.

In any case, can you verify it works for you now on a patch with [skip ci] ?

eedri:
closing, this was just for testing purposes.

PR #160, opened 2018-03-24
Fix task of remove firewalld package

gouyang:
The previously PR is placing the remove firewalld task before stop and disable firewalld, it should be placed after that task.

Signed-off-by: Guohua Ouyang <guohuaouyang@gmail.com>


gouyang:
will resend this PR on top of the previous PR

PR #165, opened 2018-03-27
storage: cns and cdi idempotency fixes

aglitke:
It should be possible to provision storage multiple times in a row
without errors.  The same is true for deprovisioning.  Make the
following changes to fix idempotency:
- cdi: ignore not found error when deleting clusterrolebinding
- cdi: Only create the clusterrolebinding if it does not exist
- cns: ignore not found error when deleting storageclass


aglitke:
Covered instead by #166

PR #168, opened 2018-03-28
cluster-login: escalate privileges when touching /etc/hosts

lukas-bednar:
In order to allow this playbook to run under non-root user, we need to explicitly ask to escalate permissions for this operation.


lukas-bednar:
@rthallisey or @gbenhaim  Could you please take a look at this change ?

lukas-bednar:
@rthallisey thanks

PR #169, opened 2018-03-28
[WIP] Storage cinder

kirankt:
The storage-cinder role deploys a self-contained Cinder storage environment consisting of a Cinder, RabbitMQ and MariaDB. MariaDB is still not fully functional. TripleO's MariaDB fails in the initContainer stage and uses ephemeral storage. We need to use a stabel PV or hostpath. We can entirely ditch TripleO's Mariadb for SCL version, if the need arises. RabbitMQ pod is not clustered and doesn't have any particular storage requirements.

This role currently supports Ceph and XtremIO backends. NetApp backend is coming soon.


playbooks/storage.yml
```diff
@@ -21,3 +21,13 @@
...
       when: storage_role == 'storage-demo'
+
+- hosts: localhost
+  vars:
+  gather_facts: False
+  # unset http_proxy. required for running in the CI
+  environment:
+    http_proxy: ""
+  roles:
+    - roles: storage-cinder
```
rthallisey: does cinder need the cdi role?

requirements.yml
```diff
@@ -4,3 +4,23 @@
   name: openshift-contrib
 
 - src: ansible.kubernetes-modules
+
+- src: https://github.com/openstack/openstack-ansible-plugins
```
rthallisey: Nice.  All these will need to be added to the kubevirt-apb's requirements.yml.

requirements.yml
```diff
@@ -4,3 +4,23 @@
...
+  version: master 
+  name: openstack-ansible-plugins
+
+- src: https://github.com/openstack/ansible-role-k8s-tripleo
```
rthallisey: This role needs ```meta/main.yaml``` or ci will fail

alexxa:
@kirankt To document a role, please use the [README.md template](./templates/roles_README.md). 

FYI, it's a good practice to title commits using the present tense and imperative mood, i.e. `Add/Update/Fix/Remove ..` not `Adding, Adds or Added ...`. Note that Git uses it too- by default it's `Update Readme`. In other words, it should correctly complete the sentence: “If accepted, this commit will <your commit message goes here>.” No punctuation needed in the end.

playbooks/storage.yml
aglitke: Yes.  All storage flavors should include it.

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,48 @@
+# Ansible Meta Role to Deploy Standalone Cinder
+
+This role aggregates Cinder, RabbitMQ and MariaDB to deploy Standalone
+Cinder with no authentication (noauth). 
+
+MariaDB is deployed on a node with label=mariadb. MariaDB uses hostPath
+for storage. 
+
+### Role Variables
+| variable | default |
```
aglitke: They have recently switched to a new standard tabular format for documenting the variables.  Please adopt it here.

roles/storage-cinder/tasks/deprovision.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Delete Service Account
+  command: kubectl delete serviceaccount {{service_account}} -n {{namespace}}
+  when: sa.stdout == ""
+
+- name: Revoke anyuid access to {{service_account}} SA if using OCP
```
aglitke: If you delete the SA doesn't this go away with it?

roles/storage-cinder/tasks/deprovision.yml
```diff
@@ -0,0 +1,36 @@
...
+  command: oc adm policy remove-scc-to-user anyuid -z {{service_account}} -n {{namespace}}
+  when: cluster == "openshift"
+
+- name: Delete {{privileged_service_account}} serviceaccount
+  command: kubectl create serviceaccount {{privileged_service_account}} -n {{ namespace }}
+  when: psa.stdout == ""
+
+- name: Revoke privileged access to {{privileged_service_account}} serviceaccount
```
aglitke: Same as above?

roles/storage-cinder/tasks/deprovision.yml
```diff
@@ -0,0 +1,36 @@
...
+  command: oc adm policy remove-scc-to-user privileged system:serviceaccount:{{ namespace }}:{{privileged_service_account}}
+  when: cluster=="openshift"
+
+- name: Select a host for Mariadb hostpath
+  command: kubectl get nodes -o jsonpath='{.items[0].status.addresses[?(@.type=="Hostname")].address}'
+  register: node_hostname
+
+- name: Label node to host Mariadb
```
aglitke: You want to remove the label right?  Also, can't you get all the nodes that have the mariadb label as a query?  Here you assume that the get nodes verb always returns them in the same order and that should not be relied upon.

roles/storage-cinder/tasks/deprovision.yml
kirankt: That is correct. Will prune it.

roles/storage-cinder/tasks/deprovision.yml
kirankt: Ack.

requirements.yml
```diff
@@ -4,3 +4,23 @@
   name: openshift-contrib
 
 - src: ansible.kubernetes-modules
+
+- src: https://github.com/openstack/openstack-ansible-plugins
+  version: master
```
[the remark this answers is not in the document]
kirankt: Ack. Going to make https://github.com/openstack/openstack-ansible-plugins:stable/queens as the default release.

roles/storage-cinder/tasks/deprovision.yml
```diff
@@ -0,0 +1,37 @@
...
+  when: psa.stdout == ""
+
+- name: Select a host for Mariadb hostpath
+  command: kubectl get nodes --selector="{{infra_node_label}}"=true
+  register: infra_node_hostnames
+
+- name: Remove label on the nodes that host Mariadb
+  command: kubectl label node "{{ item }}" "{{infra_node_label}}"-
+  with_items: "{{ infra_node_hostnames.stdout.split() }}"
+
```
alexxa: will Delete and Remove tasks pass if there is no account/label? no --ignore-errors or similar flag needed?

roles/storage-cinder/tasks/deprovision.yml
```diff
@@ -0,0 +1,37 @@
...
+- name: Include  Rabbitmq
```
alexxa: If you write RabbitMQ in README, write it that way everywhere in docstrings, please. The same for MariaDB.

roles/storage-cinder/defaults/main.yml
```diff
@@ -0,0 +1,69 @@
...
+# Array to remember enabled backends. 
+enabled_backends: []
+
+iscsi:
+    enabled: false
+
+#
+# Xtremio
+xtremio:
+    enabled: false
```
alexxa: Rename `enabled` to `prefix_enabled`, please

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,87 @@
...
+for storage.
+
+### Requirements
+
+
+### Role Variables
+| variable       | default           |choices           | comments  |
+|:-------------|:-------------|:----------|:----------|
+| action |  provision | <ul><li>provision</li><li>deprovision</li></ul>| Action to perform on the role|
+| infra_node_label | controller | | Label a node to allow MariaDB to utilize its hostpath
```
alexxa: missing closing `|`

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,87 @@
...
+| volume_backend_name | xtremio | | |
+| san_ip | | | |
+| san_login | | | |
+| san_password | | | |
+| image_volume_cache_enabled | | | |
+
+#### NetApp
+| variable       | default           |choices           | comments  |
+|:-------------|:-------------|:----------|:----------|
+| enabled | false  | <ul><li>true</li><li>false</li></ul>| Enables Netapp backend |
```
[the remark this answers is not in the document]
alexxa: I mean for exampel netapp_enabled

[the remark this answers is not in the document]
kirankt: Hi, the 'enabled' var is part of a hierarchical dict, so it will look like netapp.enabled: true when used. I'd rather just use 'enabled' than netapp_enabled.

roles/storage-cinder/tasks/deprovision.yml
kirankt: The command to remove the label will still work even though the label is missing. It will print out a message saying that the label isn't present, but the return code is still 0.

roles/storage-cinder/tasks/deprovision.yml
kirankt: Ack. Changed it for consistence. Thanks for catching this.

roles/storage-cinder/defaults/main.yml
kirankt: Same as above. The 'enabled' var is basically at <cinder-backend>.enabled. So, I'm hesitant to change it.

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,87 @@
+# Storage Cinder Role
+
+This role aggregates Cinder, RabbitMQ and MariaDB to deploy Standalone
+Cinder with no authentication (noauth).
+
+MariaDB is deployed on a node with label=mariadb. MariaDB uses hostPath
+for storage.
+
+### Requirements
```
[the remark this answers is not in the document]
kirankt: Will do. Thanks.

roles/storage-cinder/README.md
kirankt: Thank you. Fixed it.

roles/storage-cinder/defaults/main.yml
lukas-bednar: I am ok with both ways ...
```yaml
group:
    var_name: XYZ
```
either
```yaml
group_var_name: XYZ
```

On other hand I case see on line [62](https://github.com/kubevirt/kubevirt-ansible/pull/169/files/8ad35d4f677d964b7c0aa7fa2224800e55a83cec#diff-3f8e79ab052fe339f8e49f186f68f56dR62)
```yaml
netapp:
    netapp_storage_family: ontap_cluster
```
which seems to me as a duplication of group name in nested variable name.

kirankt: Those Cinder backend variables came from TripleO Cinder and Puppet Cinder projects. The 'enabled' variable is original to this role. Some backends use prefixed vars, others don't. Its not consistent.

kirankt: The 'enabled' var is referenced in many files within the role and changing it just for the sake of it makes no sense to me. Thanks.

requirements.yml
kirankt: Will do.

requirements.yml
kirankt: This repo has been moved to tripleo-apb one: https://github.com/tripleo-apb/ansible-role-k8s-tripleo

rthallisey:
this looks like it's making good progress @kirankt .  Needs a rebase and a few more comments addressed then I think we can merge.

aglitke:
The CI error is: ansible-role-k8s-rabbitmq was NOT installed successfully.  Can you please take a look at this.  Needs a rebase.

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,84 @@
...
+| ceph_authentication_type | cephx  | | |
+| ceph_mon_host | | | The IP address{es) of Ceph Monitors. Multiple IPs are specified as a comma-separated list |
+
+#### Xtremio
+| variable       | default           |choices           | comments  |
+|:-------------|:-------------|:----------|:----------|
+| enabled | false  | <ul><li>true</li><li>false</li></ul>| The var to enable or disable this backend. |
+| max_over_subscription_ratio | 40  | | |
+| use_multipath_for_image_xfer | | | |
+| volume_backend_name | xtremio | | |
```
aglitke: Any reason why this needs to be exposed?

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,84 @@
...
+| client_key | | | |
+| ceph_authentication_type | cephx  | | |
+| ceph_mon_host | | | The IP address{es) of Ceph Monitors. Multiple IPs are specified as a comma-separated list |
+
+#### Xtremio
+| variable       | default           |choices           | comments  |
+|:-------------|:-------------|:----------|:----------|
+| enabled | false  | <ul><li>true</li><li>false</li></ul>| The var to enable or disable this backend. |
+| max_over_subscription_ratio | 40  | | |
+| use_multipath_for_image_xfer | | | |
```
aglitke: Could you offer some tips on how to configure these?  If we see no reason to change them then let's not expose them for now.  Does this one have a default value?

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,84 @@
...
+| volume_backend_name | xtremio | | |
+| san_ip | | | |
+| san_login | | | |
+| san_password | | | |
+| image_volume_cache_enabled | | | |
```
aglitke: Does this have a default value?

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+#### NetApp
+| variable       | default           |choices           | comments  |
+|:-------------|:-------------|:----------|:----------|
+| enabled | false  | <ul><li>true</li><li>false</li></ul>| The var to enable or disable this backend. |
+| netapp_storage_family | ontap_cluster  | | |
```
aglitke: What are the valid options?

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,84 @@
...
+| netapp_storage_protocol: | | | |
```
aglitke: What are the valid options?  I know we do not currently support NFS so is it supposed to be iscsi?

roles/storage-cinder/README.md
```diff
@@ -0,0 +1,84 @@
...
+| nfs_shares_config | | | |
+| nfs_shares | | | |
```
aglitke: NFS is not supported.

roles/storage-cinder/defaults/main.yml
```diff
@@ -0,0 +1,69 @@
+---
+action: provision
+namespace: openstack
+service_account: cinder
+privileged_service_account: cinder-privileged
+
+storage_cinder_template_dir: "{{ role_path }}/templates"
```
aglitke: I think this is not necessary as ansible already looks in this directory for templates.

roles/storage-cinder/defaults/main.yml
```diff
@@ -0,0 +1,69 @@
...
+role_name: cinder
+
+infra_node_label: mariadb
+
+# Backends
+# Array to remember enabled backends. 
+enabled_backends: []
+
+iscsi:
+    enabled: false
```
aglitke: Is this required?

roles/storage-cinder/defaults/main.yml
aglitke: I prefer we keep it the way Kiran has it now.

roles/storage-cinder/defaults/main.yml
```diff
@@ -0,0 +1,69 @@
...
+    max_over_subscription_ratio: 40
+    use_multipath_for_image_xfer: true
+    volume_backend_name: xtremio
+    volume_driver: cinder.volume.drivers.emc.xtremio.XtremIOISCSIDriver
+    san_ip: 
+    xtremio_cluster_name:
+    san_login: 
+    san_password:
+    image_volume_cache_enabled:
```
aglitke: Any good default here?

roles/storage-cinder/defaults/main.yml
```diff
@@ -0,0 +1,69 @@
...
+    client_key: 
+    ceph_authentication_type: cephx
+    ceph_mon_host: 
+
+# Netapp
+netapp:
+    enabled: false
+    volume_driver: cinder.volume.drivers.netapp.common.NetAppDriver
+    netapp_storage_family: ontap_cluster
+    netapp_storage_protocol:
```
aglitke: default?

roles/storage-cinder/defaults/main.yml
```diff
@@ -0,0 +1,69 @@
...
+
+auth_strategy: noauth
+
+cinder_config:
+  DEFAULT:
+    public_bind_host: "0.0.0.0"
+
+role_name: cinder
+
+infra_node_label: mariadb
```
aglitke: let's make this label less likely to collide:  perhaps kubevirt.io/storage-cinder-mariadb-host

roles/storage-cinder/templates/mariadb.yml.j2
```diff
@@ -0,0 +1,48 @@
...
+        ports:
+        - containerPort: 3306
+          name: mariadb
+        volumeMounts:
+        - name: mariadb-persistent-storage
+          mountPath: /var/lib/mysql
+      volumes:
+      - name: mariadb-persistent-storage
+        hostPath:
+          path: /var/lib/mysql
```
aglitke: Should we choose a path that is less likely to collide with a db on the host?

roles/storage-cinder/README.md
kirankt: This change has been made.

rthallisey:
@kirankt can you rebase and we'll work on getting this passing ci

Issue #183, opened 2018-04-10
Some smaller issues


fabiand:
These two issues made it hard to follow the readme:

- WRT selinux it doesn't specify what "inventory" to run (e.g. let's specify it's the openshift-ansible inventory)
- The "KubeVirt parameters documentation" link is broken.

alexxa:
@fabiand Do you mean that `ansible-playbook -i inventory playbooks/selinux.yml` will not work with our [inventory](https://github.com/kubevirt/kubevirt-ansible/blob/master/inventory) and it should be one of [openshift-ansible inventories](https://github.com/openshift/openshift-ansible/tree/master/inventory)?

alexxa:
@yuvalif Please let me know if you are ok with PR #187. Thanks.

yuvalif:
commented in the review

alexxa:
Fixed in PR #187.

PR #187, opened 2018-04-11
Fix wrong link and add details on SELinux issue in Readme

alexxa:
This fixes issue #183.

### DIFF
--- a/playbooks/README.md
+++ b/playbooks/README.md
@@ -108,9 +108,12 @@ installed and a config file in ```$HOME/.kube/config```.  Also make sure you
 indentified with the cluster using ```oc login```.
 
 Before installing KubeVirt on an existing OpenShift cluster, ensure that SELinux is disabled on all hosts:
+
 ```bash
 $ ansible-playbook -i inventory playbooks/selinux.yml
 ```
+
+Be sure to update the [inventory file](../inventory) according to your OpenSift cluster configuration or use the file you used to deploy the cluster.
 
 Install KubeVirt on your OpenShift cluster:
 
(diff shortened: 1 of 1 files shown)


playbooks/README.md
```diff
@@ -108,10 +108,13 @@ installed and a config file in ```$HOME/.kube/config```.  Also make sure you
 indentified with the cluster using ```oc login```.
 
 Before installing KubeVirt on an existing OpenShift cluster, ensure that SELinux is disabled on all hosts:
+
 ```bash
 $ ansible-playbook -i inventory playbooks/selinux.yml
```
yuvalif: here you reference your inventory sample, probably better to note that this is a sample file that may be used (after being modified with the right node info), also, please note that if the user don't want to use it they may take their own inventory (otherwise, someone may just copy&paste this line as-is).

playbooks/README.md
```diff
@@ -108,10 +108,13 @@ installed and a config file in ```$HOME/.kube/config```.  Also make sure you
...
 ```
 
+Be sure to use the same inventory configuration file you used to deploy the OpenShift cluster.
```
yuvalif: assuming i used a very minimal inventory [file](https://gist.github.com/yuvalif/777b9b992cf11a3505e6577eb664357c) to install openshift. what would be missing there?

alexxa: @yuvalif Nothing. I have just tested it, works for me. Please let me know if it doesn't for you.

playbooks/README.md
alexxa: @yuvalif Will it be ok if I chage line 116 for: 

`Be sure to update the [inventory file](../inventory) according to your OpenSift cluster configuration or use the file you used to deploy the cluster.`

yuvalif:
looks good. thanks.

alexxa:
@yuvalif you are welcome, thanks for bring it up, so we can make our docs better.

PR #186, opened 2018-04-10
locate heketi namespace prior to its route

karmab:

PR #190, opened 2018-04-11
GlusterFS: Use correct namespace and auto-detect values

jarrpa:
Supersedes #186

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>


jarrpa:
@karmab This is my response to #186. Can you take a look and see if it works for you?

roles/storage-glusterfs/tasks/provision.yml
```diff
@@ -5,31 +5,28 @@
...
+    - cluster == 'openshift'
+    - heketi_url is not defined
 
 - set_fact:
     heketi_url: "{{ route_cmd.stdout }}"
   when: route_cmd is defined
 
 - name: Get heketi admin key
-  command: "oc get secret {{ heketi_secret_name }} -n {{ namespace }} -o jsonpath='{.data.key}'"
+  command: "kubectl get pod -n {{ glusterfs_namespace }} -l heketi={{ glusterfs_name_label }}pod -o jsonpath='{.spec.env[?(@.name=\"HEKETI_ADMIN_KEY\")].value}' | base64 -"
```
aglitke: In my environment (OCP 3.9) this did not work unless the jsonpath expression was defined like so:

{.items[0].spec.containers[0].env[?(@.name==\"HEKETI_ADMIN_KEY\")].value}

So you need first to index the .items array, then index into the containers array.  Also '=' needs to be changed to '=='.

roles/storage-glusterfs/tasks/provision.yml
```diff
@@ -5,36 +5,33 @@
...
```
aglitke: You're using a pipe so you need the 'shell' ansible module instead of the command module.

roles/storage-glusterfs/tasks/provision.yml
jarrpa: @rthallisey The heketi pod is part of a Deployment, which autogenerates a random pod name, hence why I search by label.

@aglitke On it.

jarrpa:
@aglitke Updated. Let me know if it works and I'll squash.

aglitke:
@rthallisey Please merge this, push to ansible-galaxy, and push a new kubevirt-apb also.

Issue #188, opened 2018-04-11
[enhancement] allow for remote execution of "oc" during install


yuvalif:
Currently [this](https://github.com/kubevirt/kubevirt-ansible/tree/master/playbooks#install-kubevirt-on-an-existing-cluster) section requires either ```oc``` to be installed on the machine that does the installation, or to do the installation from the master node.
The drawback with first option that the ```.kube/config``` has to be copied from the master to the installing machine, the drawback with the second option is that the master is used for installation, and would need git, and ansible to be installed on it. This is fine when the cluster is used by a developed (as this is already needed for using APBs in openshift), but does not make sense when the cluster is used for other purposes.
Ideally, ```oc``` is remotely executed by ansible.

rthallisey:
Looks like the kubevirt playbooks uses localhost.  I think we'll need to change it to master[0] like in storage.

yuvalif:
@dcritch is it safe to close the issue? was it addressed by PR #282 ?

dcritch:
@yuvalif Yup, this can be closed

Issue #86, opened 2018-02-21
Stop using hosts:all 


dyasny:
In several locations some plays are run against hosts: all. This presents a problem when more than just the k8s/openshift cluster is managed in the inventory, e.g. when there are ceph or gluster nodes in place or when the setup is being run on a local hypervisor and the hypervisor is in the inventory as "hypervisor" and should not be altered like the k8s cluster nodes. 

https://github.com/kubevirt/kubevirt-ansible/blob/6f32ae31095b2342f32c3f6ab64cc03a6db305a0/deploy-kubernetes.yml#L2

https://github.com/kubevirt/kubevirt-ansible/blob/0e5ae43551e1b50e569a303e9ca53e02b5705988/install-kubevirt-release.yml#L1

https://github.com/kubevirt/kubevirt-ansible/blob/2ef2c8ba752af53f9f43601dc76174649076d96e/deploy-openshift.yml#L2

https://github.com/kubevirt/kubevirt-ansible/blob/b9f4b40d2d8751e2bf82ee667fb20c71a8ea2d0a/deploy-with-lago.yml#L39

nellyc:
@dcritch please check your PR, update and merge :)

shiywang:
@nellyc guess we can close this one

PR #193, opened 2018-04-12
Targeting proper host groups in playbooks.

dcritch:
This change will narrow the scope of playbook runs to the relevant
nodes in an inventory, rather than the more broad 'all'.

This addresses issues #188, #86, #195 and #239

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -9,6 +9,7 @@ log_path = ./ansible.log
 # Ansible default path is $HOME/.ansible/cp
 # which in jenkins is 105 chars
 control_path = none
+pipelining = True
 ssh_args =
     -o CheckHostIP=no
     -o AddressFamily=inet
--- a/automation/README.md
+++ b/automation/README.md
@@ -1,5 +1,4 @@
 # Automation with oVirt CI
-
 In order to keep playbooks in this repository operational,
 it is being integrated with [oVirt CI System][ovirt-ci-system-doc].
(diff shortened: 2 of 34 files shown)


aglitke:
@copejon please review

dcritch:
Currently some playbooks expect to be run on localhost, while others are targeting the nodes. There was a previous issue to get that addressed and one playbook (storage.yml) has been merged with the switch.

Following the openshift-ansible approach, playbooks should run on the nodes/master0/etc and not localhost. It is a toss up between making it more apb compatible vs. e.g. requiring that openshift clients are installed on the ansible or tower server running the jobs.

I checked and of the 9 files changed, 3 are of the localhost->master[0] variety, while the others are going from `all` or similar to something more targeted.

I can:
1/ Remove those 3 files that are up to debate out of this PR and leave the rest until a final decision has been made
2/ Keep the PR as is, and add apb specific playbooks later
3/ Keep the PR as is, and add apb specific playbooks now.

What do you think?

dcritch:
cc @rthallisey @alexxa @lukas-bednar 
Looks like a mess due to a rebase, but I think this one is good to go?

lukas-bednar:
@dcritch could you please rebase it, that we can see what this patch change?

dcritch:
@lukas-bednar I seemed to mess up git real good, so I'm closing this request and opened up a clean one here: https://github.com/kubevirt/kubevirt-ansible/pull/282

PR #210, opened 2018-04-24
[WIP] Adding support for specifying versions of kubevirt components

dcritch:
This change will allow specifying the version number of the relevant
kubevirt images. 

It introduces a local, jinja2 version of the OpenShift template to allow for the manipulation. It may not be ideal, but I think it is the best way to handle this. unless we want to get the kubevirt folks to maintain a templated version for us. 

An alternative would be to patch/hack the kubevirt template once its retrieved, though that seems a bit messier.


dcritch:
Opened up kubevirt issue to see if we can get a solution for all groups: https://github.com/kubevirt/kubevirt/issues/966

dcritch:
Closing PR. This one is a bit of a mess.

Issue #216, opened 2018-05-06
Deprecate CI for OpenShift 3.7


cynepco3hahue:
Currently, KubeVirt developed for Kubernetes 1.9, so it does not relevant to run KubeVirt deployment for OpenShift 3.7 that uses Kubernetes 1.7(also it already has features like webhook validation that supported only from version 1.9)

PR #217, opened 2018-05-06
Deprecate Openshit 3.7 CI

gbenhaim:
KubeVirt is targeted for Kubernetes 1.9, so there is
no need to test on Openshift 3.7 that uses K8S 1.7

closes https://github.com/kubevirt/kubevirt-ansible/issues/216

### DIFF
--- a/inventory
+++ b/inventory
@@ -6,7 +6,7 @@ etcd
 [OSEv3:vars]
 openshift_deployment_type=origin
 containerized=true
-openshift_image_tag=v3.7.0
+openshift_image_tag=v3.9.0
 openshift_clock_enabled=true
 ansible_ssh_user=root
 openshift_master_identity_providers=[{'name': 'allow_all_auth', 'login': 'true', 'challenge': 'true', 'kind': 'AllowAllPasswordIdentityProvider'}]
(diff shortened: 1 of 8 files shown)


inventory
```diff
@@ -6,7 +6,6 @@ etcd
 [OSEv3:vars]
 openshift_deployment_type=origin
 containerized=true
-openshift_image_tag=v3.7.0
```
cynepco3hahue: Can you just change it to 3.9?
`openshift_image_tag=v3.9.0`

alexxa:
@gbenhaim @lukas-bednar I would remove "OpenShift 3.7" from [this line](https://github.com/kubevirt/kubevirt-ansible/blame/master/playbooks/README.md#L10) because the moment we default it to 3.9 and don't check it for 3.7, smth can stop working there. I'm just not sure if to put it in this PR or #178, I'm ok with any option. Thoughts?

inventory
gbenhaim: done

gbenhaim:
@alexxa I've fixed the doc.

alexxa:
@gbenhaim Thank you.

Issue #223, opened 2018-05-10
"deploy host as kubernetes node" task fails on non-master nodes


rwsu:
kubeadm init creates  /etc/kubernetes/pki/ca.crt on the master node. The file doesn't exist on worker nodes. The "retrieve token cat cert hash" task in the kubernetes-node role actually fails on worker nodes and the "deploy host as kubernetes node" task subsequently hangs.

```
TASK [kubernetes-node : retrieve token ca cert hash] **********************************************************
Thursday 10 May 2018  12:17:08 -0700 (0:00:17.361)       0:10:20.673 ********** 
changed: [**snip**.com] => {
    "changed": true, 
    "cmd": "openssl x509 -pubkey -in /etc/kubernetes/pki/ca.crt | openssl rsa -pubin -outform der 2>/dev/null | openssl dgst -sha256 -hex | sed 's/^.* //'", 
    "delta": "0:00:00.027171", 
    "end": "2018-05-10 19:17:01.550951", 
    "rc": 0, 
    "start": "2018-05-10 19:17:01.523780"
}

STDOUT:

e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855


STDERR:

Error opening Certificate /etc/kubernetes/pki/ca.crt
140506420115344:error:02001002:system library:fopen:No such file or directory:bss_file.c:402:fopen('/etc/kubernetes/pki/ca.crt','r')
140506420115344:error:20074002:BIO routines:FILE_CTRL:system lib:bss_file.c:404:
unable to load certificate

changed: [***snip**.com] => {
    "changed": true, 
    "cmd": "openssl x509 -pubkey -in /etc/kubernetes/pki/ca.crt | openssl rsa -pubin -outform der 2>/dev/null | openssl dgst -sha256 -hex | sed 's/^.* //'", 
    "delta": "0:00:00.027056", 
    "end": "2018-05-10 19:17:01.503563", 
    "rc": 0, 
    "start": "2018-05-10 19:17:01.476507"
}

STDOUT:

e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855


STDERR:

Error opening Certificate /etc/kubernetes/pki/ca.crt
139901588338576:error:02001002:system library:fopen:No such file or directory:bss_file.c:402:fopen('/etc/kubernetes/pki/ca.crt','r')
[...]

PR #224, opened 2018-05-11
Fix "kubeadm join" in multi node environment

rwsu:
/etc/kubernetes/pki/ca.crt resides only the master node. The previous
tasks to join nodes were failing because it was attempting to read the
missing ca.crt file.

This patch saves the join command as a fact and reads this fact on
non-master nodes to join them.

Fixes: https://github.com/kubevirt/kubevirt-ansible/issues/223


lukas-bednar:
Thank you!

PR #226, opened 2018-05-14
Adding Openshift 3.10 CI

gbenhaim:
And removing unneeded Openshift 3.7 file

### DIFF
--- a/inventory
+++ b/inventory
@@ -24,6 +24,9 @@ openshift_master_admission_plugin_config={"ValidatingAdmissionWebhook":{"configu
 # Docker
 container_runtime_docker_storage_type=overlay2
 
+# OC 3.10 node and master labeling
+openshift_node_groups=[{'name': 'node-config-infra-compute', 'labels': ['node-role.kubernetes.io/infra=true', 'node-role.kubernetes.io/compute=true']}, {'name': 'node-config-master', 'labels': ['node-role.kubernetes.io/master=true']}]
+
 # BEGIN ANSIBLE BROKER CONFIG
 openshift_hosted_etcd_storage_kind=nfs
 openshift_hosted_etcd_storage_nfs_options="*(rw,root_squash,sync,no_wdelay)"
--- a/stdci.yaml
+++ b/stdci.yaml
@@ -1,5 +1,6 @@
 substage:
   - openshift_3-9
+  - openshift_3-10
 
 runtime_requirements:
   support_nesting_level: 2
--- a/automation/check-patch.openshift_3-10.mounts
+++ b/automation/check-patch.openshift_3-10.mounts
@@ -1,0 +1,1 @@
+check-patch.mounts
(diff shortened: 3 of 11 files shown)


gbenhaim:
With https://github.com/kubevirt/kubevirt-ansible/pull/227 we can stop use the docker-storage role which requires the openshift-facts module. I hope that it will fix the issue.

gbenhaim:
I'm blocked because of missing packages in Centos repos:
https://github.com/openshift/openshift-ansible/issues/8399

rthallisey:
looks like an unset var in ci? ```Message:  last_checked_host: lago-master, last_checked_var: openshift_node_group_name;openshift_node_group_name must be defined for all nodes```

nellyc:
This is a recent change which was done in openshift-ansible - 
all nodes must have openshift_node_group_name specified during both install and upgrade.

cynepco3hahue:
Exactly
- master must have `openshift_node_group_name="node-config-master-infra"`
- node `openshift_node_group_name="node-config-compute"`
If you want that master will serve as compute node, you need to add all-in-one group name(do not remember exact name)

inventory
```diff
@@ -43,6 +43,11 @@ ansible_service_broker_etcd_image_prefix=quay.io/coreos/
 ansible_service_broker_etcd_image_tag=latest
 # END ANSIBLE BROKER CONFIG
 
+[nodes:vars]
+# Mark all nodes as infra and compute nodes (oc >= 3.10)
+openshift_node_groups=[{'name': 'node-config-infra-compute', 'labels': ['node-role.kubernetes.io/infra=true', 'node-role.kubernetes.io/compute=true']}]
```
lukas-bednar: Based on error I see in jenkins, I would try to move this `openshift_node_groups` to `[OSEv3:vars]` section.

inventory
```diff
@@ -24,6 +24,9 @@ openshift_master_admission_plugin_config={"ValidatingAdmissionWebhook":{"configu
 # Docker
 container_runtime_docker_storage_type=overlay2
 
+# OC 3.10 node and master labeling
+openshift_node_groups=[{'name': 'node-config-infra-compute', 'labels': ['node-role.kubernetes.io/infra=true', 'node-role.kubernetes.io/compute=true']}, {'name': 'node-config-master', 'labels': ['node-role.kubernetes.io/master=true']}]
```
cynepco3hahue: setting `openshift_node_groups` will set labels automatically according to the group, so you can remove it and you must have different groups, for `master`, `infra` and `compute` resources(in general every node can server as `infra` resource, but if we talk about non-HA deployment, the cluster must have only one master node)

gbenhaim:
ci test please

stdci.yaml
```diff
@@ -1,5 +1,6 @@
 substage:
   - openshift_3-9
```
nellyc: is there a point in keeping 3.9?

gbenhaim: Let's discuss it on today's meeting.
It can be removed in the following PR.

gbenhaim:
Ready to be merged.

PR #227, opened 2018-05-16
Unified LagoInitFile and update to way we deploy docker storage

gbenhaim:
#### Unified LagoInitFile:

We can use jinja to manipulate the LagoInitFile.
Removed redundant disk from the vms.

#### Configure docker storage using openshift-ansible:

Openshift ansible added support for configuring the docker storage.
Save some lines of code, and remove openshift-ansible-contrib dep.
This change also updates the storage driver to be "overlay2", which is
recommended by docker.


inventory
```diff
@@ -16,6 +16,9 @@ template_service_broker_install=false
 openshift_use_manageiq=false
 openshift_install_examples=false
 
+# Docker
+container_runtime_docker_storage_type=overlay2
```
lukas-bednar: afaik is overlay2 is not recommended for production.
I think we should stay on production side of configuration, no?

gbenhaim: I think that you refer to `devicemapper` with `loopback-lvm`

Issue #244, opened 2018-05-28
Need to add test qcow2 image for CDI e2e testing


shiywang:
I'm working on intial e2e testing on this repo, CDI support couple ways of download images:
```
func (d *dataStream) dataStreamSelector() (io.ReadCloser, error) {
	switch d.url.Scheme {
	case "s3":
		return d.s3()
	case "http", "https":
		return d.http()
	case "file":
		return d.local()
	default:
		return nil, errors.Errorf("invalid url scheme: %q", d.url.Scheme)
	}
}
```
For http without basicauth, I think we can upload a small image to this repo for e2e testing

@lukas-bednar @screeley44 @aglitke WDYT?

PR #245, opened 2018-05-28
add testvm.qcow2 image for e2e testing

shiywang:
Fixes: https://github.com/kubevirt/kubevirt-ansible/issues/244

@aglitke @screeley44 @lukas-bednar PTAL


gbenhaim:
I don't think that we should add binary file to the repo.
Can't we store the image inside a container (like in kubvirt's functional tests)?
What are the other options?

PR #255, opened 2018-06-03
Allow to run check-patch without cleanup

gbenhaim:
This option will make it very simple to run the automation
locally and investigate the env.

For deploying the env without cleanup:

./automation/check-patch.openshift_3-9.sh --skip-cleanup

For removing the env:

./automation/check-patch.openshift_3-9.sh --only-cleanup

The default is to cleanup the env.


shiywang:
I tried, script runs no problem, only my networking has some issue, so LGTM

Issue #221, opened 2018-05-09
Iptables permission denied (you must be root) when provision kubevirt-apb from OCP web console


qwang1:
This is https://github.com/ansibleplaybookbundle/kubevirt-apb/issues/54 moved here.

@rthallisey hi,

I chose ephemeral storage plan to deploy kubevirt on OCP web console, then I ran into this error when executed "Allow ceph OSD traffic" task:

iptables v1.4.21: can't initialize iptables table `filter': Permission denied (you must be root)

Here it the ansible log when deploy from web console:

```
[root@host-172-16-120-120 ~]# oc project rh-virtualization-prov-w4n5h
Now using project "rh-virtualization-prov-w4n5h" on server "https://172.16.120.120:8443".

[root@host-172-16-120-120 ~]# oc get all
NAME                                          READY     STATUS    RESTARTS   AGE
po/apb-e1ebfcc3-4a01-4c0f-83c2-8080d880c127   0/1       Error     0          1m

[root@host-172-16-120-120 ~]# oc logs po/apb-e1ebfcc3-4a01-4c0f-83c2-8080d880c127
+ [[ provision --extra-vars {"_apb_plan_id":"storage-demo","_apb_service_class_id":"60c8357b2a1cb091488d9c5586c4eb4b","_apb_service_instance_id":"94ef5eab-0670-4e69-8702-7688af1c5b0d","admin_password":"redhat","admin_user":"qwang","cluster":"openshift","namespace":"qwang-storage-demo-1","storage_role":"storage-demo","version":"0.4.1-alpha.2"} == *\s\2\i\/\a\s\s\e\m\b\l\e* ]]
+ ACTION=provision
+ shift
+ apb_action_path=kubevirt-ansible/playbooks/kubevirt.yml
+ playbooks=/etc/ansible/roles/kubevirt-ansible/playbooks/kubevirt.yml
+ CREDS=/var/tmp/bind-creds
+ TEST_RESULT=/var/tmp/test-result
+ whoami
+ '[' -w /etc/passwd ']'
++ id -u
+ echo 'apb:x:1000180000:0:apb user:/opt/apb:/sbin/nologin'
+ set +x
+ [[ -e /etc/ansible/roles/kubevirt-ansible/playbooks/kubevirt.yml ]]
+ [[ ! -d /etc/ansible/roles/kubevirt-ansible/playbooks/kubevirt.yml ]]
[...]

PR #265, opened 2018-06-07
Iptables changes need root

rthallisey:
fixes: https://github.com/kubevirt/kubevirt-ansible/issues/221


alexxa:
@rthallisey what about adding `become: yes` to one block, instead of every task?

rthallisey:
Generally, blocks are used when tasks need to be processed as a single unit; like an iteration of a loop. It's easier to add ```become: yes``` individually because these don't need to be processes as a unit.

PR #266, opened 2018-06-07
Improve performance

gbenhaim:
#### Pre pull container images on All VMs
- Pulling the containers in parallel should reduce runtime.

- Also, use Ansible "free" strategy plugin, which allows each host to run
  until the end of the play without waiting for the other hosts.

- Remove pre/post tasks, we don't use it.

#### Generate tests on the localhost
- It will be faster.

- It won't fill the master's root partition.

### DIFF
--- a/automation/check-patch.mounts
+++ b/automation/check-patch.mounts
@@ -2,3 +2,4 @@
 /var/lib/libvirt/boot
 /var/lib/lago
 /dev/shm
+/var/run/docker.sock:/var/run/docker.sock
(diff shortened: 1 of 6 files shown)


gbenhaim:
@dcritch How do you install Openshift and how much time does it take you?
I really want to reduce the time of the automation.

dcritch:
@gbenhaim I install via running openshift-ansible directly.

I took another look at this, and I'm wondering how does it effect cri-o folks. I've enabled cri-o on my last couple of rebuilds. I don't think a 'docker pull' helps. cri-o seems to do something else to get images, but it wasn't 100% clear to me as to how.

gbenhaim:
@dcritch We currently don't run with CRIO, if and when we decide to use also CRIO, this optimization can be adjusted.

PR #267, opened 2018-06-08
Remove temporary selinux playbook

rthallisey:


rthallisey:
```playbooks/cluster/openshift/config.yml``` is different.  That's needed for deploying in CI.

PR #272, opened 2018-06-10
Enable pipelining in ansible.cfg

gbenhaim:
Will reduce the number of SSH operations, thus improve
performance.


gbenhaim:
Reduce ~10 minutes!

Issue #239, opened 2018-05-27
kubevirt.yml playbook should not be hardcoded localhost


shiywang:
like storage.yml playbook change here: https://github.com/kubevirt/kubevirt-ansible/pull/173
shouldn't we keep consistency, and make host also to be master[0] or sth ?

@lukas-bednar @alexxa

shiywang:
similar to https://github.com/kubevirt/kubevirt-ansible/issues/195

lukas-bednar:
@shiywang  Actually there is localhost by design of APBs, this playbook is supposed to be executed inside of [APB container](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/design.md).

We have PR about it with some discussion going around https://github.com/kubevirt/kubevirt-ansible/pull/193#pullrequestreview-112213308

Please feel free to address your points there! Thank you.

dcritch:
Confirmed that this is addressed in PR [193](https://github.com/kubevirt/kubevirt-ansible/pull/193/)

nellyc:
@dcritch can we close this one?

dcritch:
@nellyc yup

Issue #195, opened 2018-04-13
Optimize cdi playbook


qwang1:
Hi there,
Considering situations like HA cluster, multi-node cluster, and ansible host is not one master, so I think the following modification would be better. These change can provision/deprovision CDI successfully on my cluster as well.
```
# cat kubevirt-ansible/playbooks/cdi.yml
- hosts: localhost ---> masters[0]
connection: local ---> comment it # connection: local
```

dcritch:
This will be fixed with PR [193](https://github.com/kubevirt/kubevirt-ansible/pull/193/)

PR #282, opened 2018-06-12
Targeting proper host groups in playbooks

dcritch:
This change will narrow the scope of playbook runs to the relevant
nodes in an inventory, rather than the more broad 'all'.

This addresses issues #188, #86, #195 and #239

Made a mess of the last one [0], so starting fresh 

[0]: https://github.com/kubevirt/kubevirt-ansible/pull/193


playbooks/cluster/openshift/config.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- hosts: all
+- hosts: nodes
```
gbenhaim: Should be nodes and masters (they are part of the `OSEv3` group)

playbooks/cluster/openshift/config.yml
```diff
@@ -82,7 +82,7 @@
 
 - import_playbook: "{{ openshift_ansible_dir }}/playbooks/prerequisites.yml"
 
-- hosts: all
+- hosts: nodes
```
gbenhaim: Should be nodes and masters (they are part of the `OSEv3` group)

playbooks/provider/lago/config.yml
```diff
@@ -52,7 +52,7 @@
     - name: Start Lago env
       command: lago --workdir "{{ prefix }}" start
 
-- hosts: all
+- hosts: nodes
```
gbenhaim: Should be all the VMs that lago creates, you can keep it as it is.
In ansible 2.5.3, the `all` pattern doesn't include the local host:

```
 > ansible -i /dev/null --list-hosts "all"
 [WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

  hosts (0):
```

playbooks/cluster/openshift/config.yml
dcritch: all masters are nodes

playbooks/cluster/openshift/config.yml
dcritch: all masters are nodes

playbooks/provider/lago/config.yml
dcritch: no sweat... reverting that bit

playbooks/cluster/openshift/config.yml
dcritch: I can still change them to 'masters nodes' to be more clear if you like though

playbooks/cluster/openshift/config.yml
lukas-bednar: I believe, it is not correct assumption, you can have separate master without being in `nodes` hostgroup. Please correct me if I am wrong.

dcritch: AFAIK, all masters are in the nodes group so they can share the SDN but I can't find the original source. The table from the advanced install puts the master(s) in the `nodes` group in all scenarios though.

https://docs.openshift.com/container-platform/3.9/install_config/install/advanced_install.html#adv-install-example-inventory-files

playbooks/cluster/openshift/config.yml
gbenhaim: > I can still change them to 'masters nodes' to be more clear if you like though

it will be a good solution

PR #289, opened 2018-06-14
break out docker_prefix var and update README

dcritch:
This PR splits out the docker_registry variable better to align with upstream kubevirt vs. what we want to control.

Also updated README to reflect new vars.

### DIFF
--- a/roles/kubevirt/defaults/main.yml
+++ b/roles/kubevirt/defaults/main.yml
@@ -1,6 +1,7 @@
 namespace: "kube-system"
 registry_url: "docker.io"
-docker_prefix: "{{ registry_url }}/kubevirt"
+registry_namespace: "kubevirt"
+docker_prefix: "{{ registry_url }}/{{ registry_namespace }}"
 cluster: "openshift"
 apb_action: "provision"
 release_manifest_url: "https://github.com/kubevirt/kubevirt/releases/download"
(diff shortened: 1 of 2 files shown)


roles/kubevirt/defaults/main.yml
```diff
@@ -1,6 +1,7 @@
 namespace: "kube-system"
 registry_url: "docker.io"
-docker_prefix: "{{ registry_url }}/kubevirt"
+registry_namespace: "kubevirt"
```
alexxa: I don't see `registry_namespace` variable in roles/kubevirt/README.md. 
Should it be `registry_prefix` maybe?

roles/kubevirt/defaults/main.yml
```diff
@@ -1,6 +1,7 @@
...
+docker_prefix: "{{ registry_url }}/{{ registry_namespace }}"
```
alexxa: The same here. `registry_url` maybe?

dcritch:
@alexxa nope, you're right... I typo'ed when adding to README. it should be registry_namespace in vars file and README.

I think registry_url was fine though, or maybe I don't follow.

alexxa:
@dcritch and what about updating kubevirt.yaml.j2 [L391](https://github.com/kubevirt/kubevirt-ansible/blob/master/roles/kubevirt/templates/kubevirt.yaml.j2#L391) and [L428](https://github.com/kubevirt/kubevirt-ansible/blob/master/roles/kubevirt/templates/kubevirt.yaml.j2#L428)?

dcritch:
@alexxa I'm not sure I follow. We are temporarily carrying that j2 in our repo, but is generated by kubevirt [0]. They use `{{ docker_prefix }}/virt-controller:{{docker_tag}}` but since we want to be able to specify both the registry URL and namespace/org, I exposed it as `docker_prefix: {{ registry_url }}/{{ registry_namespace }}`

[0] https://github.com/kubevirt/kubevirt/pull/1116

Issue #304, opened 2018-06-27
CI/Openshift-Ansible need to find a way to enable  DefaultStorageClass in admission controller


shiywang:
So currently CI will failed if I remove my storageclass in pvc yaml file, but according to the docs https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#enabling-dynamic-provisioning, we should be able to enable dynamic provisioning for pvc to choose the default storageclass.

So I check the code it may related to openshift-ansible repo `openshift-ansible/roles/openshift_service_catalog/templates/api_server.j2` somewhere like this place, do we have a workaround to overwrite that value in our own repo or do you think we should submit a patch to enable that in openshift-ansible ?
block https://github.com/kubevirt/kubevirt-ansible/pull/246

@gbenhaim @lukas-bednar @aglitke

shiywang:
It could also be our StorageClass yml file doesnt has `storageclass.kubernetes.io/is-default-class:true` I will  submit a pr to test that

PR #305, opened 2018-06-27
enable storageclass.kubernetes.io/is-default-class=true

shiywang:
Trying to fix https://github.com/kubevirt/kubevirt-ansible/issues/304


shiywang:
@lukas-bednar @gbenhaim @aglitke PTAL

lukas-bednar:
LGTM

aglitke:
I think the plan is to remove the kubevirt storage class and switch to everything just using the default storage class.  For CI, this should be configured in the openshift-ansible inventory along with the rest of the gluster settings.

```openshift_storage_glusterfs_storageclass_default=true```

**Comment by shiywang** [comment]
@aglitke do you mean fix it here: https://github.com/shiywang/kubevirt-ansible/blob/ba9912a408cc2825ed3a0bf727e81532233e8a55/playbooks/cluster/openshift/vars/gluster.yml#L1-L11

I guess this file will overriede the `glusterfs_storageclass_default` even we change in upstream. But seems in this file they disabled glusterfs default on purpose.

@jarrpa @rthallisey @gbenhaim

**Comment by jarrpa** [comment]
That file doesn't just disable marking the StorageClass default, it prevents the creation of the SC entirely. If you want a GlusterFS SC to be the default SC, then you'll need the following:

```
openshift_storage_glusterfs_storageclass: true
openshift_storage_glusterfs_storageclass_default: true
openshift_storageclass_default: false
```

Also, you probably don't want to be using the `gluster/heketiclone` image anymore. Unless I'm mistaken and the clone functionality isn't in mainline heketi yet.

shiywang:
@aglitke @jarrpa I think I already enable openshift glusterfs settting, but if I remove kubevirt storageclass How do we deal `external_provisioner` ?

jarrpa:
Oh, right... yeah, that does need a custom StorageClass that openshift-ansible can't provide. So you'll need to keep that.

aglitke:
Why do we need a custom external provisioner?  Without the clone PoC aren't we back to using the stock openshift gluster setup?  Thus it should be fine to simply set the openshift-provided glusterfs storage class as default and be done?

gbenhaim:
@aglitke Does this change replace `storage-glusterfs` role?

aglitke:
@gbenhaim Yes.  As far as I can tell we don't need to do anything at the kubevirt-ansible level to take advantage of the gluster provided by openshift-ansible.

shiywang:
close this since it's already merged in another pr

PR #295, opened 2018-06-20
Update KubeVirt to v0.7.0

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

PR #306, opened 2018-06-27
Update cdi version to match up with 0.7.0

rthallisey:
- cdi to 1.0
 - remove kubevirt.yaml.j2 since it's now a release artifact as of v0.7.0-alpha.5

depends-on: #295


shiywang:
This match up pr should better merged after https://github.com/kubevirt/kubevirt-ansible/pull/295

PR #314, opened 2018-07-05
Always use the j2 from kubevirt releases

rthallisey:


aglitke:
Is the CI flaky again?

rthallisey:
@aglitke I think it's ok. I accidentally closed this.

PR #317, opened 2018-07-08
Network e2e test, vms/pods connectivity 

shiywang:


gbenhaim:
ci test please

shiywang:
@nellyc hi, I think @duyanyan has no problem with this automation, do you think there's someone else need to help review this pr ?

tests/network/network_connectivity_test.go
```diff
@@ -0,0 +1,70 @@
...
+		vmbIP = vmb.Status.Interfaces[0].IP
+	})
+
+	Context("Connectivity between VMs", func() {
+
+		It("in vmb ping vma should be successful", func() {
+			expecter, err := tests.LoggedInCirrosExpecter(vmb)
+			Expect(err).ToNot(HaveOccurred())
+			defer expecter.Close()
+			cmdCheck := fmt.Sprintf("ping -w 3 %s \n", vmaIP)
```
gbenhaim: Why not check the return code?

tests/network/network_connectivity_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"kubevirt.io/kubevirt/tests"
+)
+
+var _ = Describe("Network Connectivity", func() {
+
+	flag.Parse()
+	virtClient, err := kubecli.GetKubevirtClient()
+	tests.PanicOnError(err)
+
+	var vma *v1.VirtualMachineInstance
```
gbenhaim: Please use arrays to store the VMs structs, and then replace the duplicate code with loops.

tests/network/network_connectivity_test.go
shiywang: Okay, I check the message `3 packets transmitted` so even if there's some package lost due to networking latency or sth(not 0% package loss), it still will be able to send out the package. is that Okay ?

shiywang:
@gbenhaim updated, PTAL

shiywang:
@gbenhaim PTAL, thanks

PR #318, opened 2018-07-11
Remove OC 3.9 from CI

gbenhaim:
- KubeVirt isn't supported on OC < 3.10, there is no reason to
  keep testing it in CI.

- Make 3.10 as a default for the cluster's version.


gbenhaim:
@nellyc Done.

PR #325, opened 2018-07-19
Debug

shiywang:


shiywang:
@gbenhaim do you know why testing a pr in CI spend more than 2 hours ?

gbenhaim:
@shiywang It can happen if there are no available slaves (servers).
We work on a solution that will increase the number of slaves.

PR #335, opened 2018-07-26
Set reporting style for STDCI to plain HTML

gbenhaim:
Make the links posted on tested PRs point to a static HTML file that
loads much faster then Blue Ocean.

### DIFF
--- a/stdci.yaml
+++ b/stdci.yaml
@@ -3,3 +3,6 @@ substage:
 
 runtime_requirements:
   support_nesting_level: 2
+
+reporting:
+  style: stdci


gbenhaim:
You can see the new UI in http://jenkins.ovirt.org/job/kubevirt_kubevirt-ansible_standard-check-pr/1049/artifact/ci_build_summary.html

Are you OK with this change?

lukas-bednar:
I don't really mind, it works for me too.
On other hand I liked that drop-down shells :-P

gbenhaim:
@lukas-bednar You can access the stdout of the jobs through the `Log` link, it will be much quicker than blue ocean.

If you really want blue ocean, you can click the third icon (from the right) at the top right corner.

PR #337, opened 2018-07-30
Use crio instead of docker

gbenhaim:
crio is going to be the default container runtime in OC 3.11.
Let's adopt this change in order to catch bugs early as possible.

Details:

- /var/lib/containers will be used for crio persistent data.

- Set "containerized=false", since it's not supported with crio.

- Drop "pre-pull" container images, since we install using RPMs.

### DIFF
--- a/playbooks/cluster/openshift/config.yml
+++ b/playbooks/cluster/openshift/config.yml
@@ -67,14 +67,7 @@
         - kexec-tools
         - sos
         - psacct
-        - docker
         - NetworkManager
-
-    - name: Enable and start docker service
-      service:
-        state: started
-        enabled: yes
-        name: docker
 
     - name: Enable and start NetworkManager service
       service:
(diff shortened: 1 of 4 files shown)


gbenhaim:
@cynepco3hahue any idea from which repo I can install cri-o?

cynepco3hahue:
@gbenhaim You can try to install it from https://cbs.centos.org/repos/paas7-openshift-origin310-candidate/x86_64/os/Packages/, but I can not guarantee that it compatible

playbooks/cluster/openshift/config.yml
```diff
@@ -84,14 +77,6 @@
 
 - import_playbook: "{{ openshift_ansible_dir }}/playbooks/prerequisites.yml"
 
-- hosts: masters nodes
```
lukas-bednar: you can pre-pull images too for cri-o, just use `podman` instead of `docker`

gbenhaim: Yes, but we install using rpms (`containerized=fase`)

cynepco3hahue:
Good to see that packages from CBS repo worked fine :smile:

playbooks/cluster/openshift/config.yml
cynepco3hahue: I think it still downloads some images from docker hub, worth to check.

PR #341, opened 2018-08-02
Test script for preset life-cycle testing

jerry7z:


tests/vmi_preset_lifecycle_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	"kubevirt.io/kubevirt/pkg/api/v1"
+	"kubevirt.io/kubevirt/pkg/kubecli"
+	"kubevirt.io/kubevirt/tests"
+)
+
+var _ = Describe("Preset liftcycle", func() {
+	flag.Parse()
+
+	_, err := kubecli.GetKubevirtClient()
```
gouyang: This line of code is not used, why it should be here?

tests/vmi_preset_lifecycle_test.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to write json file %s", jsonFile)
+	}
+	return jsonFile, nil
+}
+
+func runOcCreateCommand(presetJsonFile string) (string, error) {
+	out, err := tests.RunOcCommand("create", "-f", presetJsonFile)
+	return out, err
+}
```
nellyc: if we think these kind of funcs are useful (as oppose to just using RunOcCommand in the test itself, we should move them to the utils, so everybody will be able to use them

tests/vmi_preset_lifecycle_test.go
```diff
@@ -0,0 +1,171 @@
...
+				}
+				ExpectWithOffset(1, file).NotTo(BeAnExistingFile())
+			}
+		}
+
+		testGivenPreset := func() {
+
+			Context("Preset testing", func() {
+
+				It("should succeed to generate a Preset", assertGeneratedPresetJson())
```
shiywang: I dont understand why you want to incarcerate so many functions, also could you tried style like this?

```
It("should succeed to generate a Preset", func() {
          By("Generating Preset JSON file")
	  _, err := generatePresetJson(vmPreset)
	  Expect(err).ToNot(HaveOccurred())
	  Expect(presetJsonFile).To(BeAnExistingFile())
})
```
it might due to incarcerated function make the log disappeared.

tests/vmi_preset_lifecycle_test.go
```diff
@@ -0,0 +1,171 @@
...
+					err := os.Remove(file)
+					ExpectWithOffset(1, err).ToNot(HaveOccurred())
+				}
+				ExpectWithOffset(1, file).NotTo(BeAnExistingFile())
+			}
+		}
+
+		testGivenPreset := func() {
+
+			Context("Preset testing", func() {
```
shiywang: why we need this `Context` inside the function ?

tests/vmi_preset_lifecycle_test.go
jerry7z: agree, as kubevirt implement a new cmd RunCommand, i will update all these func

PR #342, opened 2018-08-02
inventory-file: remove default or unused options

lukas-bednar:
* There are options set in inventory which are defaults in
openshift-ansible repository - so no need to have them in inventory

* We don't have nfs as a part of our deployment - so no need to set nfs
options for hosted etcd storage

* Update comment about node labels

### DIFF
--- a/inventory
+++ b/inventory
@@ -7,13 +7,11 @@ etcd
 openshift_deployment_type=origin
 containerized=true
 openshift_image_tag=v3.10.0
-openshift_clock_enabled=true
 ansible_ssh_user=root
 openshift_master_identity_providers=[{'name': 'allow_all_auth', 'login': 'true', 'challenge': 'true', 'kind': 'AllowAllPasswordIdentityProvider'}]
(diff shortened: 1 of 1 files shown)


inventory
```diff
@@ -66,7 +56,6 @@ ansible_service_broker_etcd_image_tag=latest
 
 #
 # If you run openshift deployment
-# You can add your master as schedulable node with option openshift_schedulable=true
-# Add at least one node with lable to run on it router and docker containers
-# openshift_node_labels="{'region': 'infra','zone': 'default'}"
+# Add at least one node with infra annotation to run on it router and docker containers
```
alexxa: the order should be: verb (run) object (router ...) place (on ...), i.e
`to run router and docker containers on it.`
or simply:
`to run router and docker containers there.`

lukas-bednar: addressed

gbenhaim:
@lukas-bednar Ansible service broker doesn't need NFS share for etcd?

PR #347, opened 2018-08-06
This is just a test demo

shiywang:

### DIFF
--- a/tests/util.go
+++ b/tests/util.go
@@ -10,108 +10,123 @@ import (
 	. "github.com/onsi/gomega"
 
 	ktests "kubevirt.io/kubevirt/tests"
+	"github.com/davecgh/go-spew/spew"
 )
 
-type Result struct {
+type CmdBuilder struct {
+	cmd           string
 	verb          string
 	resourceType  string
 	resourceName  string
 	resourceLabel string
 	filePath      string
-	nameSpace     string
+	namespace     string
 	query         string
 	expectOut     string
 	actualOut     string
+	cmdArgs       []string
 }
-
-var KubeVirtOcPath = ""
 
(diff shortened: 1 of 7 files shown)


tests/util.go
```diff
@@ -10,108 +10,123 @@ import (
...
+		ktests.PanicOnError(fmt.Errorf("failed to write the json file %s", jsonFile))
 	}
-	return jsonFile, nil
 }
+
+func (b *CmdBuilder) Run() string {
+	var err error
+	b.actualOut, err = ktests.RunCommand(b.cmdArgs[0], b.cmdArgs[1:]...)
+	Expect(err).ToNot(HaveOccurred())
+	return b.actualOut
```
lukas-bednar: I would create CmdResult struct and set there all results (stdout, stderr, rc), define some action over that structure, like expectOutput(XY), exitsWith(RC) ... so caller of this function can control what is correct / expected result of executed command.

tests/util.go
```diff
@@ -10,108 +10,123 @@ import (
...
+	Expect(err).ToNot(HaveOccurred())
+	err = ioutil.WriteFile(jsonFile, []byte(b.actualOut), 0644)
 	if err != nil {
-		return "", fmt.Errorf("failed to write the json file %s", jsonFile)
+		ktests.PanicOnError(fmt.Errorf("failed to write the json file %s", jsonFile))
 	}
-	return jsonFile, nil
 }
+
+func (b *CmdBuilder) Run() string {
```
lukas-bednar: Might be useful to add option to pass stdin as an argument for this function

gbenhaim:
@shiywang Is this PR needed?

PR #344, opened 2018-08-03
Add node_eviction_test

gouyang:
* add test for
	- VirtualMachineInstance Eviction Behavior
	- VirtualMachineInstanceReplicaSet Eviction Behavior
	- VirtualMachine Evictions Behavior

** Notes for reviewer:**

1. It's using artifacts from kubevirt/cluster/examples instead of creating them from functions NewRandomVMI or similar ones.
2. It's using virtctl to start VM instead of using NewRepeatableVirtctlCommand.

To use the real entities looks like more a e2e test.

Signed-off-by: Guohua Ouyang <gouyang@redhat.com>

PR #350, opened 2018-08-08
Add flag kubectl-path

### LINKED ISSUE #349: Add flag kubectl-path for tests in hack/run-tests.sh
Details in http://jenkins.ovirt.org/job/kubevirt_kubevirt-ansible_standard-check-pr/1077//artifact/check-patch.openshift_3-10.el7.x86_64/mock_logs/script/stdout_stderr.log.

The tests is failed at below lines, the root cause is it cannot find kubectl binary, add `-kubectl-path=/usr/bin/kubectl` should fix it.

```
/root/go/src/kubevirt.io/kubevirt-ansible/tests/node_eviction_test.go:94
/root/go/src/kubevirt.io/kubevirt-ansible/tests/node_eviction_test.go:147
```

And it surprise me that it's not happened to oc.

gouyang:
Fixes #349

Signed-off-by: Guohua Ouyang <gouyang@redhat.com>

### DIFF
--- a/hack/run-tests.sh
+++ b/hack/run-tests.sh
@@ -8,5 +8,5 @@ prefix=kubevirt
 tag=v0.7.0
 kubeconfig=~/.kube/config
 
-${TESTS_OUT_DIR}/tests.test -kubeconfig=$kubeconfig -tag=$tag -prefix=$prefix -test.timeout 60m
-${TESTS_OUT_DIR}/cdi.test -kubeconfig=$kubeconfig -tag=$tag -prefix=$prefix -test.timeout 60m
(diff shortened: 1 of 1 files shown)


shiywang:
I've implemented a lookup function for all the commands downstream, once I update the vendor and submit a pr, it will be fixed in another different appoach, but I think this is another way to fix as well, could you make sure your pr https://github.com/kubevirt/kubevirt-ansible/pull/344 also have this code ? only add parameters here won't make the CI validate the change.

nellyc:
just wondering, why do we need kubectl?
isnt oc & virtctl enough to cover everything we need?

shiywang:
> isnt oc & virtctl enough to cover everything we need?

Yes, oc is enough from functionality perspective.

gouyang:
Change the code to use "oc adm drain" now.

PR #351, opened 2018-08-08
run-tests.sh: add FUNC_TEST_ARGS param

lukas-bednar:
Add option to pass additional arguments to test execution,
it will be useful on our environment when we want to generate junit
results for example.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -7,3 +7,6 @@ import:
 - package: kubevirt.io/kubevirt
   version: v0.7.0
   repo:    https://github.com/kubevirt/kubevirt.git
+- package: kubevirt.io/qe-tools
+  version: v0.1.2
+  repo: https://github.com/kubevirt/qe-tools.git
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -7,6 +7,8 @@ KUBEVIRT_ANSIBLE_DIR="$(
 OUT_DIR=$KUBEVIRT_ANSIBLE_DIR/_out
 TESTS_OUT_DIR=$OUT_DIR/tests
 
+FUNC_TEST_ARGS="${FUNC_TEST_ARGS:--test.timeout 60m --junit-output=junit.xml}"
+
 function build_func_tests() {
     mkdir -p ${TESTS_OUT_DIR}/
     ginkgo build ${KUBEVIRT_ANSIBLE_DIR}/tests
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -7,7 +7,8 @@ KUBEVIRT_ANSIBLE_DIR="$(
 OUT_DIR=$KUBEVIRT_ANSIBLE_DIR/_out
 TESTS_OUT_DIR=$OUT_DIR/tests
 
-FUNC_TEST_ARGS="${FUNC_TEST_ARGS:--test.timeout 60m --junit-output=junit.xml}"
+FUNC_TEST_ARGS="${FUNC_TEST_ARGS:--test.timeout 60m --junit-output=junit-func-tests.xml}"
+CDI_TEST_ARGS="${CDI_TEST_ARGS:--test.timeout 60m --junit-output=junit-cdi-tests.xml}"
 
 function build_func_tests() {
     mkdir -p ${TESTS_OUT_DIR}/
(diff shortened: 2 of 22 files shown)


lukas-bednar:
cc @nellyc

gbenhaim:
@lukas-bednar Can we add junit reports to kubevirt-ansible as well?

nellyc:
+1 @gbenhaim 
I think we can have the xunit creation enabled by default
and for our executions we will only need to add the polarion reporter

btw if xunit is not created on other projects - kubevirt, cdi - i think it is worth adding it there as well

lukas-bednar:
From some reason I have troubles to run `make deps-update` to update `glide.lock`  :-(

glide.yaml
```diff
@@ -7,3 +7,6 @@ import:
 - package: kubevirt.io/kubevirt
   version: v0.7.0
   repo:    https://github.com/kubevirt/kubevirt.git
+- package: kubevirt.io/qe-tools
+  version: 13fc8573324cd2af42c88043b5a7c6e03bf39f95
```
lukas-bednar: I just copied this from kubevirt repo, I am not sure if this is correct for us here

lukas-bednar: putting `version: v0.1.2` instead

lukas-bednar:
from some reason `--junit-output` options wasn't recognized.
```
13:39:01 [check-patch.openshift_3-10.el7.x86_64] flag provided but not defined: -junit-output
```

lukas-bednar:
Now it looks much better ...

@gbenhaim what file name pattern does junit plugin expect ? From logs I see that no junit files were collected.

hack/run-tests.sh
```diff
@@ -4,9 +4,9 @@ set -e
...
-prefix=kubevirt
-tag=v0.7.0
-kubeconfig=~/.kube/config
+prefix=${DOCKER_PREFIX:-kubevirt}
+tag=${DOCKER_TAG:-v0.7.0}
+kubeconfig=${KUBECONFIG:-~/.kube/config}
 
-${TESTS_OUT_DIR}/tests.test -kubeconfig=$kubeconfig -tag=$tag -prefix=$prefix -test.timeout 60m
-${TESTS_OUT_DIR}/cdi.test -kubeconfig=$kubeconfig -tag=$tag -prefix=$prefix -test.timeout 60m
+${TESTS_OUT_DIR}/tests.test -kubeconfig=$kubeconfig -tag=$tag -prefix=$prefix ${FUNC_TEST_ARGS}
```
lukas-bednar: It is not very good idea to share same variable for both test-suites.
For example if I pass `--junit-output=junit.xml` and pass it to both, the result will get overwritten by each other :-/

hack/common.sh
```diff
@@ -7,6 +7,9 @@ KUBEVIRT_ANSIBLE_DIR="$(
 OUT_DIR=$KUBEVIRT_ANSIBLE_DIR/_out
 TESTS_OUT_DIR=$OUT_DIR/tests
 
+FUNC_TEST_ARGS="${FUNC_TEST_ARGS:--test.timeout 60m --junit-output=exported-artifacts/tests.junit.xml}"
+CDI_TEST_ARGS="${CDI_TEST_ARGS:--test.timeout 60m --junit-output=exported-artifacts/cdi.junit.xml}"
```
nellyc: Im not sure why we need these 2 different executions,
and not just one

lukas-bednar: I don't know neither, currently result of test compilation are two test binaries ...

shiywang: it was only because in test_suit_test, `BeforeTestSuitSetup` function created some uncessary pv/pvc, sometimes create cdi pvc may failed, so I have to branchout a new TestSuit. but I have a pr to remove those things https://github.com/kubevirt/kubevirt-ansible/pull/357, but I just not sure if which way we would like to adopt.

gbenhaim:
ci test please

hack/run-tests.sh
lukas-bednar: done

lukas-bednar:
It is ready for merge, both suites are visible in jenkins : http://jenkins.ovirt.org/job/kubevirt_kubevirt-ansible_standard-check-pr/1135/testReport/(root)/

nellyc:
lgtm
it means that @shiywang's PR will need to be rebased on this and remove the reporters added to the CDI suite

lukas-bednar:
@gbenhaim @nellyc if nobody has objections I am going to merge this one today.

PR #352, opened 2018-08-09
[RFR] Added v2v tests

vatsalparekh:
New tests for V2V

### DIFF
--- a/tests/manifests/import-vm-v2v-apb.yml
+++ b/tests/manifests/import-vm-v2v-apb.yml
@@ -1,0 +1,44 @@
+---
+apiVersion: v1
+kind: Template
+metadata:
+  name: v2v-template
+  annotations:
+    description: "Template for v2v import from VMWare"
+objects:
+- apiVersion: servicecatalog.k8s.io/v1beta1
+  kind: ServiceInstance
+  metadata:
+    name: v2v-${APB_PLAN_NAME}
+  spec:
+    clusterServiceClassExternalName: ${APB_NAME}
+    clusterServicePlanExternalName: ${APB_PLAN_NAME}
+    parameters:
+      admin_user: ${OCP_ID}
+      admin_password: ${OCP_PASS}
+      url: ${VMWARE_URL}
+      vm_name: ${VM_NAME}
+      username: ${VMWARE_ID}
+      password: ${VMWARE_PASS}
+      os_type: ${V2V_OS_TYPE}
+parameters:
+- name: APB_NAME
+  description: name of the apb
+  value: dh-import-vm-apb
+- name: APB_PLAN_NAME
+  description: name of the plan in apb
+- name: OCP_ID
+  description: OCP username
+- name: OCP_PASS
+  description: password of the OCP user
+- name: VMWARE_URL
+  description: VMWare url for v2v
+- name: VM_NAME
+  description: name of the vm to import
+- name: VMWARE_ID
+  description: username of VMWare
+- name: VMWARE_PASS
+  description: password of VMWare user
+- name: V2V_OS_TYPE
+  description: type of the OS
+  value: linux
--- a/tests/v2v_test.go
+++ b/tests/v2v_test.go
@@ -1,0 +1,62 @@
+package tests_test
+
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	"kubevirt.io/kubevirt-ansible/tests"
+)
+
+// template parameters
+const (
+	v2vServiceInstance    = "tests/manifests/import-vm-v2v-apb.yml"
+	dstv2vServiceInstance = "/tmp/test-v2v.json"
+)
+
+// Array of enviroment variables to get from testing env
+var listofEnvVars = []string{"ocpID", "ocpPass", "VMWareURL", "VMName", "VMWareID", "VMWarePass"}
+
+// Map to store Enviroment variables
+var v2vEnvVars map[string]string
+
+// Get all the Env variable needed for tests
+func getV2vEnv() {
+	v2vEnvVars = map[string]string{}
+	for _, vars := range listofEnvVars {
+		val, ok := os.LookupEnv(vars)
+		// skip the vars that are not found
+		if ok {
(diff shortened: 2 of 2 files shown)


tests/v2v/v2v_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/kubevirt/kubevirt-ansible/tests"
+	. "github.com/onsi/ginkgo"
+
+	ktests "kubevirt.io/kubevirt/tests"
+)
+
+var _ = Describe("Importing and Starting VMs via V2V", func() {
+	flag.Parse()
+	ktests.SkipIfNoOc()
+
```
nellyc: we need some skip for the case there is no vmware instance to use

nellyc:
ci test please

vatsalparekh:
@nellyc @shiywang @gouyang PTAL
v2v tests would fail atm because of errors in v2v itself
but test does create serviceinstance so that's working

tests/v2v_test.go
```diff
@@ -0,0 +1,55 @@
...
+const (
+	v2vServiceInstance    = "tests/manifests/import-vm-v2v-apb.yml"
+	dstv2vServiceInstance = "/tmp/test-v2v.json"
+)
+
+// Array of enviroment variables to get from testing env
+var listofEnvVars = []string{"ocpID", "ocpPass", "VMWareURL", "VMName", "VMWareID", "VMWarePass"}
+
+// Map to store Enviroment variables
+var v2vEnvVars map[string]string
```
shiywang: I believe you can also use struct and iteratively fetch the value like this: http://www.golangprograms.com/how-to-get-struct-variable-information-using-reflect-package.html, But I'm okay with `map` as well.

tests/v2v_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Get all the Env variable needed for tests
+func getV2vEnv() {
+	v2vEnvVars = map[string]string{}
+	for _, vars := range listofEnvVars {
+		val, ok := os.LookupEnv(vars)
+		if !ok {
+			//raise exception
```
shiywang: raise exception?

shiywang:
others LGTM

tests/v2v_test.go
vatsalparekh: the env vars are important for the test, so `log.fatal` becomes important

tests/v2v_test.go
vatsalparekh: I feel `map` is fine here as storing them in a loop while taking input from array is easy and works

tests/manifests/import-vm-v2v-apb.yml
```diff
@@ -0,0 +1,38 @@
...
+  spec:
+    clusterServiceClassExternalName: ${APB_NAME}
+    clusterServicePlanExternalName: vmware
+    parameters:
+      admin_user: ${OCP_ID}
+      admin_password: ${OCP_PASS}
+      url: ${VMWARE_URL}
+      vm_name: ${VM_NAME}
+      username: ${VMWARE_ID}
+      password: ${VMWARE_PASS}
```
pkliczewski: What about `os_type`? Do we want to have it here?

tests/manifests/import-vm-v2v-apb.yml
```diff
@@ -0,0 +1,38 @@
...
+  annotations:
+    description: "Template for v2v import from VMWare"
+objects:
+- apiVersion: servicecatalog.k8s.io/v1beta1
+  kind: ServiceInstance
+  metadata:
+    name: import-vm-apb
+  spec:
+    clusterServiceClassExternalName: ${APB_NAME}
+    clusterServicePlanExternalName: vmware
```
pkliczewski: What about `vmware-template`? Do we want to verify it as well?

tests/manifests/import-vm-v2v-apb.yml
vatsalparekh: included

tests/manifests/import-vm-v2v-apb.yml
vatsalparekh: yes, included

tests/manifests/import-vm-v2v-apb.yml
vatsalparekh: but it will only be useful when testing windows VMs

pkliczewski: correct

tests/v2v_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+var _ = Describe("Importing VMs using V2V", func() {
+
+	BeforeEach(func() {
+		getV2vEnv()
+	})
+
+	Context("Create ServiceInstance of import-vm-apb with vmware plan", func() {
+		It("should import vm suceessfully", func() {
+			tests.ProcessTemplateWithParameters(v2vServiceInstance, dstv2vServiceInstance,
```
shiywang: 44-52 and 59-67 are exact same which can be reduced by either puting those code into a `BeforeEach` or `JustBeforeEach` block or using `table.Entry`

vatsalparekh: https://github.com/kubevirt/kubevirt-ansible/blob/b5e629df47c2d733f54c426052f072eab07f4c4e/tests/v2v_test.go#L45
https://github.com/kubevirt/kubevirt-ansible/blob/b5e629df47c2d733f54c426052f072eab07f4c4e/tests/v2v_test.go#L60
@shiywang the param `APB_PLAN_NAME` is different in both, so it is not repeating same thing
Can you suggest a way to parametrize it?

shiywang: can't you put those two into `v2vEnvVars`? if not another way is define a local var like `apbPlanName`, then both will be  `"APB_PLAN_NAME="+apbPlanName,`

vatsalparekh: @shiywang @nellyc added use of Ginkgo table to parametrize tests from single Describe

tests/v2v_test.go
lukas-bednar: As we have discussed these config options will move to config file in future.

vatsalparekh:
ci test please

PR #354, opened 2018-08-11
add expose ssh, console, vnc tests for vm/vmi

shiywang:

### DIFF
--- a/tests/expose_test.go
+++ b/tests/expose_test.go
@@ -1,0 +1,98 @@
+package tests_test
+
+
+import (
+	"flag"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ktests "kubevirt.io/kubevirt/tests"
+	"github.com/davecgh/go-spew/spew"
+	"kubevirt.io/kubevirt-ansible/tests"
+	"kubevirt.io/kubevirt/pkg/kubecli"
+)
+
+const vmName = "vm-cirros"
+
+var _ = Describe("D1", func() {
+
+	flag.Parse()
+	virtClient, err := kubecli.GetKubevirtClient()
+	ktests.PanicOnError(err)
+
+	ktests.SkipIfNoOc()
+
+	Context("C1", func() {
+
+		It("the VM should be created successfully", func() {
+			tests.CreateResourceWithFilePathTestNamespace("tests/manifests/vm-cirros.yml")
+		})
+
+		It("the VM should be started by virtctl", func() {
+			args := []string{"start", vmName, "-n", tests.NamespaceTestDefault}
+			_, err := ktests.RunCommand("virtctl", args...)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		Specify("the VMI should be running", func() {
+			tests.WaitUntilResourceReadyByNameTestNamespace("vmi", vmName, "-o=jsonpath='{.status.phase}'", "Running")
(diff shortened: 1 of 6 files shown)


shiywang:
This is not ready for review!!!

gbenhaim:
@shiywang Can you please add reviewers?

shiywang:
@gbenhaim I don't have permission to add reviewers here. But I guess I would ping @fabiand to help me assign reviewers ?

gouyang:
The title says "... vnc tests", but I don't see vnc tests in it.

tests/tests_suite_test.go
```diff
@@ -15,9 +17,10 @@ func TestTests(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	tests.BeforeTestSuitSetup()
+	tests.LookupBinaryCommand()
```
gouyang: I don't think it should be here. If it fails, it prevent other tests to run.

shiywang: No, because it shouldn't be failed. and this function is bump from upstream since it removed it.

gouyang: Why it shouldn't be failed?

gouyang: Take a look at `sanity_test.go`, it does not use any of these binaries. If this check failed, it prevent that tests to be run. And not all of them are required for a single tests, like `kubectl` is not necessary for most of the case.

gbenhaim:
@shiywang  @gouyang  Is this PR ready?

shiywang:
@nellyc Anyone else needs to have a look on this one?

tests/tests_suite_test.go
shiywang: I remove this function `LookupBinaryCommand` already

tests/expose_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+		It("the VM should be exposed by virtctl via ClusterIP", func() {
+			args := []string{"expose", "vmi", vmName, "-n", tests.NamespaceTestDefault, "--name=myvm-ssh", "--port=22", "--type=ClusterIP"}
+			_, err := ktests.RunCommand("virtctl", args...)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("the VMI should be able to connect via ssh", func() {
+			By("Create a pod as ssh client to connect to VMI")
+			args := []string{"run", testPod, "-n", tests.NamespaceTestDefault, "--image=shiywang/test-ssh", "--", "sleep", "3600"}
```
nellyc: what's shiywang/test-ssh?

gbenhaim:
@shiywang @nellyc Any update on this one?

tests/expose_test.go
nellyc: ping

tests/expose_test.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+var _ = Describe("VM/VMI system test", func() {
+
+	flag.Parse()
+	virtClient, err := kubecli.GetKubevirtClient()
+	ktests.PanicOnError(err)
+
+	Context("Expose service via ssh and console VNC", func() {
+
```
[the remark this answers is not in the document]
nellyc: ping

tests/expose_test.go
shiywang: it's because the testing vm was exposed by ClusterIP, which either you execute ssh on the node, or start a pod to access it, so shiywang/test-ssh is the pod I created to acesss VM via ssh, I believe we can move this to kubevirt repo, but I don't have any permissions

PR #359, opened 2018-08-20
add JustBeforeEach to CDI tests

shiywang:

### DIFF
--- a/tests/cdi/cdi_test.go
+++ b/tests/cdi/cdi_test.go
@@ -1,61 +1,50 @@
 package cdi_test
 
 import (
-	"flag"
 	"os"
 
 	. "github.com/onsi/ginkgo"
 	"kubevirt.io/kubevirt-ansible/tests"
-	ktests "kubevirt.io/kubevirt/tests"
 )
 
 var _ = Describe("Importing and starting a VM using CDI", func() {
-	flag.Parse()
-	ktests.SkipIfNoOc()
-	dstPVCFilePath := "/tmp/test-pvc.json"
-	dstPVCFilePath1 := "/tmp/test-pvc1.json"
-	dstVMFilePath := "/tmp/test-vm.json"
+	var dstPVCFilePath, dstVMFilePath, newPVCName, url string
 
-	url, ok := os.LookupEnv("STREAM_IMAGE_URL")
-	if !ok {
-		url = pvcEPHTTPNOAUTHURL
-	}
+	BeforeEach(func() {
+		var ok bool
+		dstPVCFilePath = "/tmp/test-pvc.json"
+		dstVMFilePath = "/tmp/test-vm.json"
+		newPVCName = pvcName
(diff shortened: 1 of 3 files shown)


tests/cdi/cdi_test.go
```diff
@@ -13,49 +13,39 @@ var _ = Describe("Importing and starting a VM using CDI", func() {
...
 	}
 
-	Context("PVC with valid image url will succeed", func() {
-		It("should create the PVC successfully", func() {
-			tests.ProcessTemplateWithParameters(rawPVCFilePath, dstPVCFilePath, "PVC_NAME="+pvcName, "EP_URL="+url)
-			tests.CreateResourceWithFilePathTestNamespace(dstPVCFilePath)
-		})
+	BeforeEach(func() {
+		tests.ProcessTemplateWithParameters(rawPVCFilePath, dstPVCFilePath, "PVC_NAME="+pvcName, "EP_URL="+url)
+	})
```
gouyang: The `tests.ProcessTemplateWithParameters` runs here and invokes in the 2nd context again? I think it does not make sense, how about put it into `JustBeforeEach`?

shiywang: updated

shiywang:
@nellyc @gbenhaim @alexxa @lukas-bednar @aglitke  PTAL

tests/cdi/cdi_test.go
```diff
@@ -1,61 +1,50 @@
...
 
-	Context("PVC with valid image url will succeed", func() {
-		It("should create the PVC successfully", func() {
-			tests.ProcessTemplateWithParameters(rawPVCFilePath, dstPVCFilePath, "PVC_NAME="+pvcName, "EP_URL="+url)
-			tests.CreateResourceWithFilePathTestNamespace(dstPVCFilePath)
-		})
+	Context("PVC with valid image url", func() {
 
-		Specify("the PVC should become bound", func() {
+		It("will succeed", func() {
```
nellyc: I liked it more before when we had more info on the TC - 
instead of removing the specify & it, 
could we change them to 'by'?
this way it will be like test steps and not different test cases

shiywang: Sorry, I'm not quit understand what do you mean here, I don't think we could change them to `by`?
so previously I made each step a `It` block, but after I read the documentation carefully, I found out `BeforeEach` and `JustBeforeEach` is two good functions for us to decouple the configuration and resource creation, and improve the code readability. `It` block as a minimal execution unit run in between `BeforeEach` `JustBeforeEach` and `AfterEach`.

nellyc: as you pointed out today, 'by' is actually in the func call under util.go, so that should be enough
(even though this means we wont be able to parse it in a way which will make the steps appear in Polarion this way)

nellyc:
can you please rebase?

shiywang:
@nellyc done, PTAL

PR #323, opened 2018-07-18
kubevirt: fix condition error on render offline template task

guits:
this condition might fail if the task which registers `offline_templates` is skipped.

Typical error:

```
The conditional check 'offline_templates.stat.exists == True' failed.
The error was: error while evaluating conditional
(offline_templates.stat.exists == True): 'dict object' has no attribute
'stat'
```

The idea is to check if `offline_templates` is skipped before trying to
evaluate `offline_templates.stat.exists`

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #361, opened 2018-08-22
Remove OC 3.9 from CI (#318)

lukas-bednar:
- KubeVirt isn't supported on OC < 3.10, there is no reason to
  keep testing it in CI.

- Make 3.10 as a default for the cluster's version.

Signed-off-by: gbenhaim <galbh2@gmail.com>

PR #365, opened 2018-08-27
kubevirt: fix condition error on render offline template task (#323)

rthallisey:
this condition might fail if the task which registers `offline_templates` is skipped.

Typical error:

```
The conditional check 'offline_templates.stat.exists == True' failed.
The error was: error while evaluating conditional
(offline_templates.stat.exists == True): 'dict object' has no attribute
'stat'
```

The idea is to check if `offline_templates` is skipped before trying to
evaluate `offline_templates.stat.exists`

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>
(cherry picked from commit e9bc98023313f788dcc3db9c6f4c4cadeff43b14)

### DIFF
--- a/roles/kubevirt/tasks/provision.yml
+++ b/roles/kubevirt/tasks/provision.yml
@@ -43,7 +43,7 @@
   template:
     src: "{{ offline_template_dir }}/v{{ version }}/kubevirt.yaml.j2"
     dest: "/tmp/kubevirt.yaml"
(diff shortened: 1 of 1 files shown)


lukas-bednar:
That 3.9 has to go away ... I have PR for that here #361 but it fails on missing virtualization ability which is weird ... :-/

PR #372, opened 2018-09-04
Add Datavolume E2E test

shiywang:


tests/cdi/cdi_suite_test.go
```diff
@@ -13,14 +13,20 @@ import (
...
+	vmAPIVersion = "kubevirt.io/v1alpha2"
+)
+
+const (
+	rawDataVolumePath = "tests/manifests/datavolume.yml"
+	rawPVCFilePath    = "tests/manifests/golden-pvc.yml"
+	rawVMFilePath     = "tests/manifests/test-vm.yml"
+)
+
+const (
```
jeffvance: nit and it's a matter of personal style, but these 3 constants blocks could be combined.

tests/cdi/cdi_test.go
```diff
@@ -2,51 +2,101 @@ package cdi_test
...
+	"k8s.io/apimachinery/pkg/util/rand"
 	"kubevirt.io/kubevirt-ansible/tests"
+	ktests "kubevirt.io/kubevirt/tests"
 )
 
-var _ = Describe("Importing and starting a VM using CDI", func() {
-	var dstPVCFilePath, dstVMFilePath, newPVCName, url string
+var _ = Describe("Importing and starting a VMI using CDI", func() {
+	prepareCDIResource := func(manifest, url string) string {
+		dstFilePath := tmpTestDir + "test-prepare-file-" + rand.String(5) + ".json"
```
jeffvance: consider using the `filepath` pkg and calling Join to build the pathname. This is more portable across different OSs

tests/cdi/cdi_test.go
```diff
@@ -2,51 +2,101 @@ package cdi_test
...
+	waitForImporterPodWriteImg := func(phase, resourceName string) {
+		switch phase {
+		case "Succeeded":
+			tests.WaitUntilResourceReadyByNameTestNamespace("pvc", resourceName, "-o=jsonpath='{.metadata.annotations}'", "pv.kubernetes.io/bind-completed:yes")
+			tests.WaitUntilResourceReadyByLabelTestNamespace("pod", tests.CDI_LABEL_SELECTOR, "-o=jsonpath='{.items[*].status.phase}'", "Succeeded")
+		case "Failed":
+			tests.WaitUntilResourceReadyByLabelTestNamespace("pod", tests.CDI_LABEL_SELECTOR, "-o=jsonpath='{.items[*].status.phase}'", "Failed")
+		case "NOExist":
+			Consistently(func() bool {
+				args := []string{"get", "pods", "-l", tests.CDI_LABEL_SELECTOR, "-n", tests.CDI_LABEL_SELECTOR}
```
jeffvance: so you've defined that the pod's namespace is the same value as it's label?

tests/cdi/cdi_test.go
```diff
@@ -2,51 +2,101 @@ package cdi_test
...
 
-	Context("PVC with invalid image url", func() {
-		BeforeEach(func() {
-			newPVCName = pvcName1
-			url = invalidPVCURL
-		})
+	AfterEach(func() {
+		By("Deleting all pvc with the oc-delete command")
+		args := []string{"delete", "pvc", "--all", "-n", tests.NamespaceTestDefault}
+		_, err := ktests.RunCommand("oc", args...)
```
jeffvance: no label selector needed here?

tests/util.go
```diff
@@ -39,6 +40,15 @@ const (
 	paramFlag            = "-p"
 )
 
+func CreateDirIfNotExist(dir string) {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		err = os.MkdirAll(dir, 0755)
+		if err != nil {
+			panic(err)
```
jeffvance: generally don't call Panic in utility funcs. If Panic must be called then consider changing the func name to something like `CreateDirIfNotExistOrDie`

tests/util.go
```diff
@@ -39,6 +40,15 @@ const (
 	paramFlag            = "-p"
 )
 
+func CreateDirIfNotExist(dir string) {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
```
jeffvance: I believe this call to Stat is not needed because `MkDirAll` does not return an error if the dir already exists.

tests/util.go
```diff
@@ -92,6 +102,11 @@ func CreateResourceWithFilePathTestNamespace(filePath string) {
 	execute(Result{cmd: "oc", verb: "create", filePath: filePath, nameSpace: NamespaceTestDefault})
 }
 
+func DeleteResourceWithNameTestNamespace(resourceType, resourceName string) {
```
jeffvance: it would be a bit more general if the ns is also passed here. You could decide that if ns == "" to use the test default.

tests/cdi/cdi_suite_test.go
shiywang: Actually I personally prefer use compact style, so I think you need to change back your code @gouyang wdyt ?

gouyang: @shiywang I don't understand the question for me? Could you explain a bit more?

shiywang: it means instead of split const
```
const (
a = 1
b = 2
) 

const (
c = 3
)
```
we prefer
```
const (
a = 1
b = 2
c = 3
)
```

shiywang:
@jeffvance PTAL

tests/cdi/cdi_suite_test.go
gouyang: Group variables and constants are an idiom in go. I don't understand why you're going to anti it.

shiywang:
ci test please

PR #373, opened 2018-09-04
Install packages using the native packages manager

gbenhaim:
The "package" module doen't have the same optimization as the "yum:
module when installing multiple packages.

Since we are not testing on Fedora images, there is no need to also use the
"dnf" moudle.


gbenhaim:
Reduce ~30 sec

PR #379, opened 2018-09-07
Add tests for vm templates

gouyang:
Signed-off-by: Guohua Ouyang <gouyang@redhat.com>


shiywang:
this PR is similar to my https://github.com/kubevirt/kubevirt-ansible/pull/354, we can figure it out how to combine those two

tests/util.go
```diff
@@ -32,11 +32,15 @@ type Result struct {
 var KubeVirtOcPath = ""
 
 const (
-	CDI_LABEL_KEY        = "app"
-	CDI_LABEL_VALUE      = "containerized-data-importer"
-	CDI_LABEL_SELECTOR   = CDI_LABEL_KEY + "=" + CDI_LABEL_VALUE
+	TemplateNS           = "openshift"
```
shiywang: why you need to create another namespace, is there anything you cant do in `kubevirt-test-default` ?

gouyang: The vm templates are residing in ns 'openshift', we need to test it to ensure they're working.

gouyang: It

> this PR is similar to my #354, we can figure it out how to combine those two

There are would be some tests are similar at the setup, it would be okay.
We should not combine different tests for different purposes.

gouyang:
> this PR is similar to my #354, we can figure it out how to combine those two

There are would be some tests are similar at the setup, it would be okay.
We should not combine different tests for different purposes.

shiywang:
@gszasz could you please take a look of this one ?

tests/template_test.go
```diff
@@ -0,0 +1,91 @@
+package tests_test
+
```
nellyc: I think the file name is confusing, as it is not testing template functionality, but the existing templates,
i suggest to rename to something like template_example_tests

gszasz: @nellyc In CNV we want to be sure we can use Template mechanism to create fully fledged 'VirtualMachine` object that passes a consequent set of `VirtualMachine` tests. Since @gouyang also verified if the Template parameters were propagated correctly into the `VirtualMachine` object properties. This alone should be sufficient to verify the template functionality from the CNV perspective (although I would use Cirros image here rather than Fedora to speed up the test a bit). When we already know that Template mechanism is functional, we can be applied the same set of tests to all our example templates, to test the templates. Since the test aimed to test both the Template functionality (from the CNV perspective) and the example templates, I would not see it confusing.

tests/template_test.go
```diff
@@ -0,0 +1,91 @@
...
+			dir, err := ioutil.TempDir("", "vm-template-")
+			Expect(err).ToNot(HaveOccurred())
+
+			vm.Manifest = filepath.Join(dir, "vm-template-fedora.yaml")
+			vm.Name = "fedora"
+			vm.Template = "vm-template-fedora"
+			vm.Params = []string{"-p", "NAME=" + vm.Name, "-p", "CPU_CORES=2", "-p", "MEMORY=2048Mi"}
+
+			if _, err := vm.ProcessClusterTemplate(); err != nil {
+				Skip(fmt.Sprintf("error occurrs when process template %s in namespace %s", vm.Template, tests.TemplateNS))
```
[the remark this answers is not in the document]
gouyang: It looks `Skipf` is not defined in ginkgo when I try to use it.

gouyang:
@gszasz @nellyc Thanks for the review and the very detailed comments. I've changed accordingly.

1. Add `annotaion` to `Expect(err)`.
2. Add constants `DefaultTimeoutForVMReady` and `DefaultPollInterval`, so other tests can use them as well.
3. `Skipf` is not working for me. 
4. Only skip the test by `template xxx could not be found`, other errors in `oc process` should be treating as failed tests.
5. Use `fedora` because it's testing `vm-template-fedora`, it's not testing the general template functionalities, it's testing the templates shipped/installed in the cluster. Because these templates are there, so we need to ensure they're working.
6. Rename file to `template_examples_test`.

Additional notes:
The templates are not installed in d/s cnv 1.2, if there is no templates in the cluster, the PR is no needed.

tests/template_test.go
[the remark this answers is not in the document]
gszasz: Although Ginkgo defines `Skipf` in the interface, it indeed does not implement the function yet. I was mistaken because `Skipf` is often used by the Kubernetes test framework. It turned out that Kubernetes defines it as an [util function](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/util.go):

```go
func Skipf(format string, args ...interface{}) { 
    msg := fmt.Sprintf(format, args...)
    log("INFO", msg)
    ginkgowrapper.Skip(nowStamp() + ": " + msg)
} 
```

tests/template_test.go
```diff
@@ -0,0 +1,91 @@
...
+			_, err = ktests.RunCommand("virtctl", args...)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Check that the virtualmachine instance is running")
+			args = []string{"get", "-n", ns, "vmi", vm.Name, "--template", "{{.status.phase}}"}
+			Eventually(func() string {
+				output, err := ktests.RunCommand("oc", args...)
+				Expect(err).ToNot(HaveOccurred())
+				return output
+			}, time.Minute*10).Should(Equal("Running"))
```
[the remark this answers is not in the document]
gszasz: Alternatively, you can also set the default values via Gomega functions. That comes handy if you need to redefine defaults for the whole test suite or a portion of the test suite.

```go
SetDefaultEventuallyTimeout(10 * time.Minute)
SetDefaultEventuallyPollingInterval(2 * time.Second)
```

PR #380, opened 2018-09-10
decrease PVC requests storage to make sure pass on localstorage

shiywang:
Signed-off-by: shiywang <shiywang@redhat.com>


gbenhaim:
@shiywang Is this still needed?

shiywang:
@gbenhaim Yes, somethings 10Gi will make our d/s env out of storage, so I think 1G or 500Mi is enough

PR #393, opened 2018-10-01
Introduce playbook for Kubevirt UI installation

mareklibra:
Playbook to install [Kubevirt Web UI](https://github.com/kubevirt/web-ui).

Based on openshift-console playbook from openshift-ansible.

Verified on OCP 3.11 and 3.10


roles/kubevirt_web_ui/tasks/start.yml
```diff
@@ -0,0 +1,46 @@
...
+  # https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment
+  command: >
+    {{ openshift_client_binary }} rollout status deployment/console --config={{ openshift.common.config_base }}/master/admin.kubeconfig -n kubevirt-web-ui
+  changed_when: false
+  # Ignore errors so we can log troubleshooting info on failures.
+  ignore_errors: yes
+  register: console_rollout_status
+
+# Log the result of `oc status`, `oc get pods`, `oc get events`, and `oc logs deployment/console` for troubleshooting failures.
+- when: console_rollout_status.rc != 0
```
rthallisey: One concern I have with debugging inside the playbook is if a failure occurs in a debug task.  Even though you have ```ignore_errors: true```, a task can still log a failure, which can be misleading.

Instead of doing the debugging, you can print out a useful error message  then recommend  commands to debug.

roles/kubevirt_web_ui/tasks/install.yml
```diff
@@ -0,0 +1,131 @@
...
+
+# kubevirt-ansible: yedit from lib_utils (openshift-ansible) is not present, so rewritten
+# Generate a new config when a config map is not defined.
+
+- name: Set web console config properties from inventory variables
+  block:
+    - name: Change consoleBaseAddress
+      lineinfile:
+        path: "{{ mktemp.stdout }}/{{ __console_config_file }}"
+        regexp: "^  consoleBaseAddress:.*$"
```
rthallisey: Could use a jinja2 template?

roles/kubevirt_web_ui/tasks/install.yml
```diff
@@ -0,0 +1,131 @@
...
+- name: Check if kubevirt-web-ui project exists
+  shell: "{{ openshift_client_binary }} projects -q | grep -w kubevirt-web-ui | awk '{ print $1 }'"
+  register: ns
+
+- name: Create kubevirt-web-ui project
+  shell: "{{ openshift_client_binary }} new-project kubevirt-web-ui"
+  when: ns.stdout != "kubevirt-web-ui"
+
+- name: Make temp directory for console templates
+  command: mktemp -d /tmp/console-ansible-XXXXXX
```
rthallisey: ansible may have a built in command

roles/kubevirt_web_ui/tasks/install.yml
mareklibra: Sounds reasonable considering the code in isolation.
But this change would complicate things when merging with potentially existing ConfigMap. More importantly, I would farther deviate from the openshift-ansible implementation.

Please note, the `kubevirt-web-ui` is a fork of `openshift/console`. For that reason, I would like to stay aligned with their playbook (lives in `openshift-ansible`) as much as possible - it will simplify maintenance in the future.

We work hard to keep both upstream and the fork in sync, as they might be merged in the future. The kubevirt-ansible playbook introduces nothing new compared to the openshift-ansible, just adapts to different project environment.

roles/kubevirt_web_ui/tasks/install.yml
mareklibra: Found this one: https://docs.ansible.com/ansible/2.5/modules/tempfile_module.html

Would be simpler change than with the jinja2 template, but again: the role will be harder to compare with its openshift-ansible counterpart.

Can we keep it as it is? From semantical POV, there would be no difference.

roles/kubevirt_web_ui/tasks/install.yml
rthallisey: Let's move it to it's own file (`render-config.yaml`) so it's easier to track.  Then you can call the `render-config.yaml` tasks from `install.yaml`.  It will make this more readable so it's easier to maintain.

roles/kubevirt_web_ui/tasks/install.yml
rthallisey: will leave as is

roles/kubevirt_web_ui/tasks/install.yml
mareklibra: @rthallisey , so the task has been moved to `console-config.yml`.

mareklibra:
@rthallisey , can this be merged, please?

rthallisey:
needs a rebase @mareklibra.  Sorry I lost track of it

PR #409, opened 2018-10-14
Increase Gluster's deployment timeout

gbenhaim:
Once in a while the deployment fails because of a timeout (which in most
cases is a false positive). Let's increase the timeout.

```release-note
None
```


gbenhaim:
ci test please

gbenhaim:
ci test please

Issue #385, opened 2018-09-12
should have option to pass custom timeout


vatsalparekh:
https://github.com/kubevirt/kubevirt-ansible/blob/8d91e5da424729865847d5520f482e2e64754c60/tests/util.go#L147

tests such as v2v #352 might require more then 2m to appear, having a param to pass custom waiting time would be helpful in such cases

shiywang:
yes, we can do that, I'll add it in a folloow up pr

shiywang:
fixed in https://github.com/kubevirt/kubevirt-ansible/pull/384 
@vatsalparekh

PR #411, opened 2018-10-15
add framework dir change file structure and add a customized timeot function

shiywang:

Fixes https://github.com/kubevirt/kubevirt-ansible/issues/385

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -1,15 +1,14 @@
 package tests_test
 
 import (
-	"os"
-
 	. "github.com/onsi/ginkgo"
-	"kubevirt.io/kubevirt-ansible/tests"
+	f "kubevirt.io/kubevirt-ansible/tests/framework"
+	"time"
 )
 
 // template parameters
 const (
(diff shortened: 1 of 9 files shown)


shiywang:
@nellyc this is only some file structure change and add a customized timeout function to fix https://github.com/kubevirt/kubevirt-ansible/issues/385

shiywang:
@nellyc do you know who can help review this?

shiywang:
@aglitke @jeffvance

shiywang:
Hi, @lukas-bednar I totally agree what you suggested, the only problem is that would be a huge change, especially for `Framework struct there holding some configuration parameters which affect functions executions ` part, I have to think how to abstract those common used steps into Framework struct and let people to comfortable with how to use it, so it will take some time but each time rebase will be a pain in the ass.

So do you think we can break it down into 2-3 separated prs? first I will move files into framework package and break down functions into different files to make them more readable as you suggested, then I will submit a follow-up pr to address how we abstract functions into Framework struct, wdyt?

My first pr is here, PTAL: https://github.com/kubevirt/kubevirt-ansible/pull/473

PR #414, opened 2018-10-17
Increase Nodes RAM

gbenhaim:
We keep adding more components to the deployment,
which cause the nodes RAM run out.
Adding 1GB of RAM to each node (except from the master which already
has 4GB). Now each node has 3GB of RAM.

```release-note
None
```


lukas-bednar:
Looks good, lets see if it helps !

PR #417, opened 2018-10-19
Reduce required parameters for kubevirt-web-ui playbook

mareklibra:
**What**: If not set by the user, gather the `public_master_hostname` and `openshift_master_default_subdomain` parameters automatically

With this change, the only required attribute is the `kubevirt_web_ui_image_name` (docker image).

The values of `public_master_hostname` and `openshift_master_default_subdomain` are gathered only if not provided by the user explicitly.
These fact defaults are gathered from openshift-console deployment, assuming it's namespace and URL compositions as hardcoded within openshift-ansible. If this assumption is not met, the fallback value is empty string ("").

**Release note**:

```release-note
The "public_master_hostname" and "openshift_master_default_subdomain" parameters of the "kubevirt-web-ui" playbook are optional now. If not provided explicitly, they are determined from the existing openshift-console deployment.
```

### DIFF
--- a/roles/kubevirt_web_ui/defaults/main.yml
+++ b/roles/kubevirt_web_ui/defaults/main.yml
@@ -1,23 +1,14 @@
 ---
-# Required param:
-# - kubevirt_web_ui_image_name
-#
-# TODO: gather automatically
-# - openshift_master_default_subdomain
-# - public_master_hostname
+openshift_client_binary: "oc"
 
 kubevirt_web_ui_nodeselector: {"node-role.kubernetes.io/master":"true"}
 
 __console_template_file: "console-template.yaml"
 __console_config_file: "console-config.yaml"
 
-# Provide image explicitely
-# kubevirt_web_ui_image_name: "docker.io/mareklibra/kubevirt-web-ui:f679e704219f58aea97a1433ea01e7c7227afc7d"
-
 # Default the replica count to the number of masters.
 kubevirt_web_ui_replica_count: "1"
 
-kubevirt_web_ui_hostname: "kubevirt-web-ui.{{openshift_master_default_subdomain}}"
 kubevirt_web_ui_cert: ""
 kubevirt_web_ui_key: ""
 kubevirt_web_ui_ca: ""
@@ -26,15 +17,6 @@ kubevirt_web_ui_auth_ca_file: "/var/run/secrets/kubernetes.io/serviceaccount/ca.
 
 kubevirt_web_ui_documentation_base_url: "https://docs.openshift.com/container-platform/3.11/"
 
-######
-# Mimic openshift-ansible
-openshift_client_binary: "oc"
-
 openshift:
   common:
-    config_base: "/etc/origin"  # as hardcoded in playbooks/cluster/openshift/config.yml
-  master:
-    public_console_url: "https://{{ public_master_hostname }}/console" # Example: https://master.mycluster.com:8443/console
-    public_api_url: "https://{{ public_master_hostname }}" 
-    logout_url: "" # TODO: not needed so far
-
+    config_base: "/etc/origin" # as hardcoded in playbooks/cluster/openshift/config.yml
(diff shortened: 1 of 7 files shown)


mareklibra:
cc @rthallisey

mareklibra:
cc @rthallisey

mareklibra:
cc @rthallisey

mareklibra:
cc @rthallisey

mareklibra:
cc @rthallisey

mareklibra:
cc @rthallisey

mareklibra:
cc @rthallisey

mareklibra:
cc @rthallisey

roles/kubevirt_web_ui/defaults/main.yml
```diff
@@ -1,23 +1,14 @@
 ---
-# Required param:
-# - kubevirt_web_ui_image_name
-#
-# TODO: gather automatically
-# - openshift_master_default_subdomain
-# - public_master_hostname
+openshift_client_binary: "oc"
```
rthallisey: why make this a var?

roles/kubevirt_web_ui/tasks/required_params.yml
```diff
@@ -0,0 +1,38 @@
...
+#### Autodiscovery of missing variables
+# Assumptions for autodiscovery:
+# - console's namespace is hardcoded to "openshift-console" (as done in openshift-ansible)
+# - the openshift console is mandatory part of OKD installation
+
+- name: Discover public_master_hostname from openshift console deployment
+  shell: "{{ openshift_client_binary }} get configmap console-config -n openshift-console -o jsonpath='{.data}' | grep 'masterPublicURL:' | sed 's/^.*masterPublicURL: *http.*:\\/\\///g'"
+  register: parsedMasterPublicURL
+  when: public_master_hostname | default("") | string == ""
+
```
rthallisey: can any of these be defaults? Or are they dynamic in the sense that a static default won't always be accurate?

mareklibra: I don't know about any reasonable static default for these two variables (`public_master_hostname` or `openshift_master_default_subdomain`). The default in examples states sort of `foo.bar.example.com`. 

Recent implementation either
- gathers the values from console's deployment (should be nearly 100% cases, if I don't miss something)
- or fails on undefined `public_master_hostname` or `openshift_master_default_subdomain`

I can add check later to this file for the `undefined` value of these two facts so we fail early and with more explanatory message (means pointing to documentation). But this flow is not expected to happen and is covered in documentation. WDYT?

roles/kubevirt_web_ui/defaults/main.yml
mareklibra: The line has been moved from bellow, this change is left-over from development of this PR - not needed anymore but causing no harm.
The variable itself is taken from the original openshift-ansible, kept here for consistency between these two projects. Considering rest of the kubevirt-ansible code, it has no benefit since the `oc` is referenced here directly.

roles/kubevirt_web_ui/tasks/required_params.yml
rthallisey: My thinking is that these vars can be set as required vars and we fail without them being set instead of going out to find them.   But, if you think that the configmap will always be there and it will always use the same names to look up these vars, then it could work.

mareklibra: In my understanding, the configmap, resp. `openshift-console`, will be there as it is (recently) mandatory part of an openshift 3.11 installation. The key names of the configmap are not expected to be changed but we can keep monitoring such potential change as the kubevirt/web-ui is fork of the openshift/console and we regularly keep it in sync.
So if we can gather these values automatically, I would not delegate that on the user.

If the assumption above is not met, the execution is expected to fail instead of passing corrupted values.

mareklibra:
Rebased only.

rthallisey:
@gbenhaim is this a known issue in CI?

mareklibra:
If the issue is unrelated, can we land this PR, please?
It's required by #418 and the #418 is required by 1.3 release.

rthallisey:
Failing CI blocks merging.  I can override it, but I want to make @gbenhaim is aware.

gbenhaim:
@rthallisey CI was fixed.

PR #422, opened 2018-10-24
dockerized: Use whitelist for rsync

gbenhaim:
Copy to the build container only the files that are needed
for the test compilation.

```release-note
None
```


hack/dockerized.sh
```diff
@@ -56,7 +56,17 @@ _rsync() {
 }
 
 # Copy kubevirt into the persistent docker volume
-_rsync --delete --exclude _out --exclude deployment-openshift ${KUBEVIRT_ANSIBLE_DIR}/ "rsync://root@127.0.0.1:${RSYNCD_PORT}/build"
+_rsync \
+    --delete \
```
cynepco3hahue: hm what about `Makefile`?

gbenhaim: Done

PR #438, opened 2018-11-04
Skydive fork

leongold:
What this PR does / why we need it:
Allow the user to deploy skydive

```release-note:
Deploy skydive-project into the cluster
```


leongold:
ci test please

gbenhaim:
@leongold I've rebased this PR on https://github.com/kubevirt/kubevirt-ansible/pull/439 
I want to make sure that the test order is the cause for the failures.

leongold:
@gbenhaim, CI seems to be stuck at "waiting for status to be reported"? :(

gbenhaim:
ci test please

gbenhaim:
A rebase solve the issue.
We dropped the requirement of being able to merge only if the PR is rebased on master. Looks like it confused Github and the request for starting a Jenkins job wasn't sent.

SchSeba:
ci test please

roles/skydive/tasks/deprovision.yml
```diff
@@ -0,0 +1,14 @@
...
+
+- name: Delete {{ skydive_namespace }} namespace on the kubernetes cluster
+  shell: kubectl delete namespaces {{ skydive_namespace }}
+  when: ns.stdout == skydive_namespace
+    and cluster=="kubernetes"
+
+- name: Delete {{ skydive_namespace }} project on the openshift cluster
+  shell: oc delete namespace {{ skydive_namespace }}
+  when: ns.stdout == skydive_namespace
+    and cluster=="openshift"
```
SchSeba: please add a new line

gbenhaim:
@leongold Can you please rebase? I've sent a PR which should fix the timeout issues.

leongold:
Hurray! looks ready? @gbenhaim @SchSeba

roles/skydive/tasks/deprovision.yml
```diff
@@ -0,0 +1,14 @@
...
+  shell: "kubectl get ns | grep -w {{ skydive_namespace }} | awk \'{ print $1 }\'"
+  register: ns
+
+- name: Delete {{ skydive_namespace }} namespace on the kubernetes cluster
+  shell: kubectl delete namespaces {{ skydive_namespace }}
+  when: ns.stdout == skydive_namespace
+    and cluster=="kubernetes"
+
+- name: Delete {{ skydive_namespace }} project on the openshift cluster
+  shell: oc delete namespace {{ skydive_namespace }}
```
SchSeba: please change the namespace to project.
`oc delete project`

leongold: Done

leongold:
@gbenhaim @SchSeba looks ready?

roles/skydive/tasks/provision.yml
```diff
@@ -0,0 +1,72 @@
+---
+- name: Login As Super User
```
rthallisey: this task is no longer needed since login is always done at the playbook level

roles/skydive/tasks/provision.yml
```diff
@@ -0,0 +1,72 @@
...
+- name: Check if namespace "{{ skydive_namespace }}" exists
+  shell: "kubectl get ns | grep -w {{ skydive_namespace }} | awk \'{ print $1 }\'"
+  register: ns
+
+- name: Create {{ skydive_namespace }} namespace on the kubernetes cluster
+  shell: kubectl create namespace {{ skydive_namespace }}
+  when: ns.stdout != skydive_namespace
+    and cluster=="kubernetes"
+
+- name: Create {{ skydive_namespace }} project on the openshift cluster
```
rthallisey: this task is redundant.  `kubectl create namespace` will accomplish the same thing

roles/skydive/tasks/provision.yml
```diff
@@ -0,0 +1,72 @@
...
+    dest: /tmp/skydive.yaml
+  when: cluster=="kubernetes"
+
+- name: Render skydive deployment yaml for openshift
+  template:
+    src: openshift-skydive.yaml
+    dest: /tmp/skydive.yaml
+  when: cluster=="openshift"
+
+- name: Deploy skydive for openshift
```
rthallisey: this is redundant with the below task.  You can do an apply with both clients.

SchSeba: I think you can't because the openshift yaml is:

```yaml
apiVersion: v1
kind: Template
metadata:
  name: skydive
...
```
If you try to run `kubectl apply -f` on this file it will fail.

roles/skydive/tasks/provision.yml
SchSeba: I think It's not the same because here the project is related to the user that makes the login into the cluster.

roles/skydive/tasks/provision.yml
SchSeba: Good to know! I will open a PR to remove this from the multus deployment also.

leongold: Done

roles/skydive/tasks/provision.yml
rthallisey: Projects are namespaces++.  They have some resource constraint capabilities and a few other things.  If you explicitly are using something that a project offers out of the box, then it's not redundant.  Most of the time, namespaces will cover what you need, so I wanted to make sure.

roles/skydive/tasks/provision.yml
rthallisey: What's the purpose of creating the template?  Why not use the same yaml?

leongold:
@gbenhaim failures seem unrelated :( https://jenkins.ovirt.org/job/kubevirt_kubevirt-ansible_standard-check-pr/1694//artifact/check-patch.openshift_3-10.el7.x86_64/tests.junit.xml

roles/skydive/tasks/provision.yml
SchSeba: I used the supported deployment for every environment by the skydive team

openshift (https://github.com/skydive-project/skydive/tree/master/contrib/openshift)
kubernetes (https://github.com/skydive-project/skydive/tree/master/contrib/kubernetes)

We can use the kubernetes one for both, I just wanted to use the official way for every one of the platforms.

rthallisey: Using a template without variables and yaml resources separated with `---` are identical.  Looks like there's a difference between the two with resources.  Let's keep as is.

PR #443, opened 2018-11-05
Increase timeout when validating oc cmd results

gbenhaim:
We saw some sporadic failures on CI due to timeouts.
Increase the default timeout and allow to run oc cmd with
a custom timeout.

```release-note
None
```

### DIFF
--- a/tests/util.go
+++ b/tests/util.go
@@ -45,6 +45,11 @@ const (
 	paramFlag          = "-p"
 )
 
+const (
+	ShortTimeout       = time.Duration(2) * time.Minute
+	LongTimeout        = time.Duration(4) * time.Minute
+)
+
 func CreateNamespaces() {
 	virtCli, err := kubecli.GetKubevirtClient()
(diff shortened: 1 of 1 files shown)


tests/util.go
```diff
@@ -45,6 +45,11 @@ const (
 	paramFlag          = "-p"
 )
 
+const (
+	ShortTimeout       = time.Duration(2) * time.Minute
```
SchSeba: This variable is not used, all the tests have a timeout of 4 min now.
Maybe it will be better to change the execute function to receive a timeout variable and configure the timeout on every test to get a better control over the test run time.

gbenhaim: I agree with you, but I would like to go with this approach in a following PR.
Currently, every second PR is failing because of a timeout, and we can't  this situation.

lukas-bednar:
LGTM

tests/util.go
SchSeba: Sounds good!

PR #444, opened 2018-11-06
Collect obj description on test failure

gbenhaim:
Collect the description of the requested objects type, from all
of the projects, on test failure. The following is an example of
the generated file tree:

exported-artifacts/
├── obj-desc
│   └── after
│       └── rbac_test.go-52
│           ├── default
│           │   ├── node
│           │   │   ├── lago-master
│           │   │   ├── lago-node0
│           │   │   └── lago-node1
│           │   └── pod
│           │       ├── docker-registry-1-hlwv2
│           │       ├── registry-console-1-ns4n9
│           │       ├── router-1-2mtvx
│           │       └── router-1-5srgx
│           ├── glusterfs
│           │   └── pod
│           │       ├── glusterfs-storage-x5vk5
│           │       ├── glusterfs-storage-zphsm

...

* The number in the test file name represents the line that caused the
  failure.

```release-note
None
```

### DIFF
--- a/tests/util.go
+++ b/tests/util.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -36,6 +38,7 @@ const (
 	UsernameTestUser     = "kubevirt-test-user"
 	NamespaceTestDefault = "kubevirt-test-default"
 	UsernameAdminUser    = "test_admin"
+	PasswordAdminUser    = "123456"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


tests/util.go
```diff
@@ -273,3 +276,160 @@ func OpenConsole(virtCli kubecli.KubevirtClient, vmiName string, vmiNamespace st
...
+		td.LineNumber,
+	)
+	dest := filepath.Join("exported-artifacts", "obj-desc", "after", testName)
+
+	admin := User{Name: UsernameAdminUser, Password: PasswordAdminUser}
+	err := admin.Login()
+	if err != nil {
+		fmt.Printf(
+			"Failed to login as Admin user. Skipping obj desc collection",
+		)
```
lukas-bednar: You are probably missing `return` here, otherwise in case of failure it will proceed further.

gbenhaim:
An example of the output https://jenkins.ovirt.org/job/kubevirt_kubevirt-ansible_standard-check-pr/1700/artifact/check-patch.openshift_3-10.el7.x86_64/obj-desc/after/dummy_test.go-16/

tests/util.go
gbenhaim: Fixed

PR #446, opened 2018-11-07
Replace cluster extra var with platform

rthallisey:
OpenShift removed the endpoint that identified it was openshift.  The Broker
doesn't seem to accurately identify which cluster it's running on so let's replace
the check for openshift.

```release-note
NONE
```


rthallisey:
@lukas-bednar this PR fixed all the issue I had in the apb

roles/cdi/defaults/main.yml
```diff
@@ -2,7 +2,7 @@ cdi_namespace: "{{ cdi_image_namespace | default('kube-system') }}"
 cdi_storageclass: "{{ cdi_kubevirt_storageclass | default('kubevirt') }}"
 apb_action: "provision"
 docker_prefix: "{{ cdi_repo_tag | default('kubevirt') }}"
-docker_tag: "{{ cdi_release_tag | default('v1.2.0') }}"
+cdi_release_tag: "v1.2.0"
```
lukas-bednar: I believe we should use cdi 1.3

nellyc: correct. @shiywang already has a PR for it

rthallisey:
Seems like the cluster deployment failed? @gbenhaim 
```
Error from server (NotFound): users.user.openshift.io "test_admin" not found
```

gbenhaim:
@rthallisey Adding:

```
   environment:
     http_proxy: ""
```

to the play will solve your problem. Without it `oc` tries to use the proxy.

rthallisey:
ci test please

gbenhaim:
@rthallisey 

```
{"command":"/usr/bin/oc get vmi vmi-with-sidecar-hook -o=jsonpath='{.status.phase}' -n kubevirt-test-default -n kubevirt-test-default","component":"","level":"error","msg":"command failed: cannot run command","output":"","pos":"utils.go:2004","reason":"exit status 1","stderr":"error: the server doesn't have a resource type \"vmi\"\n","timestamp":"2018-11-08T19:36:58.230841Z"}
```

How can it be?

lukas-bednar:
@gbenhaim kubevirt wasn't deployed or kubevirt containers didn't rollout probably ?

gbenhaim:
ci test please

gbenhaim:
The node that run `virt-api` got `System OOM encountered`, might be related.
I'll post a patch to increase the memory of the nodes.

gbenhaim:
@shiywang I think that the assertion in the second tests in `e2e_test` has a bug:

You can see the importer pod description:

https://jenkins.ovirt.org/job/kubevirt_kubevirt-ansible_standard-check-pr/1721/artifact/check-patch.openshift_3-10.el7.x86_64/obj-desc/after/e2e_test.go-59/kubevirt-test-default/pod/importer-golden-pvc1-cbntk/*view*/

The pod is running, but it's in a `CrashLoopBackOff`.

@rthallisey Since this is the only test the fails and if @shiywang approve that the test has a bug, I think that we can skip CI.

rthallisey:
Merging this and we can sort out the CI issue later. Failure doesn't look related.

rthallisey:
ci test please

PR #457, opened 2018-11-13
CDI: small fix to cdi provision tasks

tiraboschi:
Fix CDI provision from an ansible based operator:
- add a missing default value
- execute an oc specific command only if oc is available


tiraboschi:
@rthallisey @irosenzw can you please review this?

shiywang:
ci test please

PR #461, opened 2018-11-14
Reduce the max and min heap on the elastic search process.

SchSeba:
Reduce the max and min heap on the elastic search process.

**Release note**:

```release-note
none
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt-ansible/461)


SchSeba:
@gbenhaim I think it's ready can you please check the logs?

PR #467, opened 2018-11-14
Lowercase bools since we're doing a string comparision

rthallisey:
Some when statements were being ignored because of case.  Use lowercase everywhere, which is consistent with the string return from `stat`.
-->
```release-note
None
```

### DIFF
--- a/roles/kubevirt/tasks/provision.yml
+++ b/roles/kubevirt/tasks/provision.yml
@@ -29,25 +29,25 @@
   stat:
     path: "{{ offline_template_dir }}/v{{ version }}/kubevirt.yaml.j2"
   register: offline_templates
-  when: byo_template.stat.exists == False
+  when: byo_template.stat.exists == false
 
 - name: Download KubeVirt Template
   get_url:
     url: "{{ release_manifest_url }}/v{{ version }}/kubevirt.yaml.j2"
     dest: "{{ kubevirt_template_dir }}/kubevirt.yaml.j2"
-  when: byo_template.stat.exists == False and offline_templates.stat.exists == False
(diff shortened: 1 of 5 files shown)


roles/kubevirt_web_ui/tasks/provision.yml
```diff
@@ -10,7 +10,7 @@
 - name: Make temp directory for console templates
   command: mktemp -d /tmp/console-ansible-XXXXXX
   register: mktemp
-  changed_when: False
+  changed_when: false
```
lukas-bednar: was this one intended ? because here you don't compare strings.

roles/kubevirt/tasks/provision.yml
```diff
@@ -65,25 +65,25 @@
...
         dest: "/tmp/demo-content.yaml"
-      when: offline_demo_content is not skipped and offline_demo_content.stat.exists == True
+      when: offline_demo_content is not skipped and offline_demo_content.stat.exists == true
 
     - name: Copy BYO Demo Content to /tmp
       copy:
         src: "{{ kubevirt_template_dir }}/demo-content.yaml"
         dest: "/tmp/demo-content.yaml"
-      when: byo_demo_content.stat.exists == False and (offline_demo_content is skipped or offline_demo_content.stat.exists == False)
+      when: byo_demo_content.stat.exists == false and (offline_demo_content is skipped or offline_demo_content.stat.exists == false)
```
lukas-bednar: are you sure that it is string ? I can read in documentation that it is boolean https://docs.ansible.com/ansible/2.7/modules/stat_module.html#return-values

?

shiywang:
ci test please

PR #469, opened 2018-11-15
Set flaky CDI test in pending mode

gbenhaim:
"PVC with invalid image url" test from e2e test is
failing even when the importer pod is failing.
The reason is that we try to assert that the pod is in Failed state,
while OC keep restarting it, so the status is Running.

```release-note
None
```


lukas-bednar:
@shiywang once you are back could you please stabilize this test?

gbenhaim:
`sidecar` is the only test that fails (and should be in another PR). Skipping CI.

PR #483, opened 2018-11-23
Add regression tests: create VMs from templates, control it via virtctl

dshchedr:
**What this PR does / why we need it**: Adds regression tests

Tests covers: 
1) Create/Delete VM via oc and kubectl
2) Control VM via virtctl (start, connect to console, check IP, stop)
3) Negative test:  delete non existing VM

Fixes #

 Now it works for Fedora only. In plans: add rhel7 and windows

**Release note**:

```release-note
NONE
```


tests/regression_functional_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+						_, err = expecter.ExpectBatch([]expect.Batcher{
+							&expect.BSnd{S: "ifconfig\n"},
+							&expect.BExp{R: ipAddress},
+						}, 120*time.Second)
+						Expect(err).ToNot(HaveOccurred())
+					}
+				})
+
+				It("Stop and delete " + osName + " VM'", func() {
```
nellyc: this creates a dependency between tests
i'd prefer to see it as 1 test

tests/regression_functional_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+					By("Start VM")
+					_, _, err := tests.StartStopVirtualMachineVirtctl(osName, "start")
+					Expect(err).ToNot(HaveOccurred())
+
+					By("Waiting until VMI is running")
+					tests.WaitUntilResourceReadyByNameTestNamespace("vmi", osName, "-o=jsonpath='{.status.phase}'", "Running")
+				})
+
+				It("Stop and delete " + osName + " VM'", func() {
```
nellyc: same

tests/regression_functional_test.go
```diff
@@ -0,0 +1,122 @@
...
+					By("Waiting until VMI is running")
+					tests.WaitUntilResourceReadyByNameTestNamespace("vmi", osName, phaseQuery, "Running")
+
+					ipAddress := tests.GetVirtualMachineSpecificParameters("vmi", osName, ipQuery)
+					// removing quotes
+					ipAddress = ipAddress[1 : len(ipAddress)-1]
+
+					By("Waiting for the console and check IP address (now only for Fedora)")
+					if osName == "testvm-fedora" {
+						expecter, err := tests.LoggedInFedoraExpecter(osName, tests.NamespaceTestDefault, 300)
```
ILpinto: Please create generic exepctor that will handle all images (can use switch case https://tour.golang.org/flowcontrol/9)

tests/regression_functional_test.go
```diff
@@ -0,0 +1,122 @@
...
+const (
+	temporaryJson = "/tmp/tmp-vm.yml"
+	ipQuery = "-o=jsonpath='{.status.interfaces[0].ipAddress}'"
+	phaseQuery = "-o=jsonpath='{.status.phase}'"
+)
+
+
+var _ = Describe("Regression and Functional tests of VMs and VMIs", func() {
+
+	flag.Parse()
```
petrkotas: Why are you parsing command line flags? I do not see it used anywhere.

tests/regression_functional_test.go
petrkotas: We are not using the generic expecter, because it would be too complicated to write one. Each VM requires different steps to get log in. This approach is fine.

tests/regression_functional_test.go
petrkotas: I agree, running It inside the loop is not good practice. Test should be predictable.

tests/regression_functional_test.go
```diff
@@ -0,0 +1,122 @@
...
+			Context("Create/Delete " + osName + " VM from template via oc command", func() {
+
+				It("Create " + osName + " VM, start and check status", func() {
+					By("Create VM")
+					tests.ProcessTemplateWithParameters(templatePath, temporaryJson, "NAME="+osName, "CPU_CORES=2")
+					tests.CreateResourceWithFilePathTestNamespace(temporaryJson)
+
+					By("Start VM")
+					_, _, err := tests.StartStopVirtualMachineVirtctl(osName, "start")
+					Expect(err).ToNot(HaveOccurred())
```
petrkotas: Please, when writing Expect, add label to it `Expect(err).ToNot(HaveOccurred(), "should start vm")` Without the label there is no information, when the test fails.

tests/regression_functional_test.go
ILpinto: the test is run with parameter from CLI [1], so we need to parse them
[1]
${TESTS_OUT_DIR}/tests.test -kubeconfig=$kubeconfig -tag=$tag -prefix=$prefix -oc-path=${OC_PATH} -kubectl-path=${KUBECTL_PATH} -virtctl-path=${VIRTCTL_PATH} ${FUNC_TEST_ARGS}

tests/manifests/vm-template-fedora.yml
```diff
@@ -0,0 +1,67 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
dshchedr: The template virt-testing-vm.yml creates VMI, but I need to create VM (to verify functionality of kubevirt (start/stop VM). In my opinion it's better to have two separate templates: one for VM and one for VMI.
I created another template with name: vm-template-common.yml. In future, we can use it for running windows and rhel VMs.

tests/regression_functional_test.go
dshchedr: I agree.
I combined this two tests

tests/regression_functional_test.go
dshchedr: Tests combined runs in loop (for for oc and kubectl)

tests/regression_functional_test.go
dshchedr: changed to using "switch". If OS is not supported - console would not be checked.

tests/regression_functional_test.go
dshchedr: Agree! labels  added

PR #484, opened 2018-11-26
add IOthread - first test case

sHaggYcaT:

Implements first test case for IOthread feature (check shared IOthread policy) 

**Release note**:

```Other test cases for IOthread feature comming soon  
```

### DIFF
--- a/tests/iothreads_test.go
+++ b/tests/iothreads_test.go
@@ -1,0 +1,206 @@
+/*
+ * This file is part of the KubeVirt project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2018 Red Hat, Inc.
+ *
+ */
+
+package tests_test
+
+import (
+	"flag"
+	"os/exec"
+	"time"
+	"encoding/xml"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
+	"kubevirt.io/kubevirt/pkg/api/v1"
+	"kubevirt.io/kubevirt/pkg/kubecli"
+	"kubevirt.io/kubevirt/tests"
+)
+
+var _ = Describe("IOThreads", func() {
+	flag.Parse()
+
+	virtClient, err := kubecli.GetKubevirtClient()
+	tests.PanicOnError(err)
+
+	gracePeriod := int64(30)
+	sharedPolicy := v1.IOThreadsPolicyShared
+	_true := true
+	IOThreadSpec := v1.VirtualMachineInstanceSpec{
+	    TerminationGracePeriodSeconds: &gracePeriod,
+	    Domain: v1.DomainSpec{
+		Resources: v1.ResourceRequirements{
+		    Requests: k8sv1.ResourceList{
+			k8sv1.ResourceMemory: resource.MustParse("1024M"),
+			k8sv1.ResourceCPU: resource.MustParse("2"),
+		    },
+		},
+		IOThreadsPolicy: &sharedPolicy,
+
+		Devices: v1.Devices{
+		    Disks: []v1.Disk{
+			{
+			    Name:       "ded1",
+			    VolumeName: "ded1volume",
+			    DiskDevice: v1.DiskDevice{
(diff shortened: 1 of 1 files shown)


tests/iothreads_test.go
```diff
@@ -0,0 +1,206 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"kubevirt.io/kubevirt/pkg/virt-launcher/virtwrap/api"
+	"kubevirt.io/kubevirt/pkg/api/v1"
+	"kubevirt.io/kubevirt/pkg/kubecli"
+	"kubevirt.io/kubevirt/tests"
```
petrkotas: I personally do not like the idea, you are importing our tests. These can change and will break your code.

tests/iothreads_test.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+var _ = Describe("IOThreads", func() {
+	flag.Parse()
```
petrkotas: Why are you parsing command line flags in this test?

tests/iothreads_test.go
```diff
@@ -0,0 +1,206 @@
...
+	Context("IOThreads Policies", func() {
+
+		// CNV-1046
+		It("Virtual Disk Settings - IOThreads", func() {
+
+			// How many symbols in the names of pod and VMI shoul match
+			symbols_to_compare := 30
+
+			IOThreadVMI, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(IOThreadVMI)
+			// Step 1
```
petrkotas: I see you are annotating the test. Ginkgo offers the way of annotating by using `By`. Example
```
By("Creating the virtual machine")
```
When using `By` you gain the benefit of seeing progress when the `ginkgo -v` is used.

tests/iothreads_test.go
```diff
@@ -0,0 +1,206 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(IOThreadVMI.Name) > symbols_to_compare).To(BeTrue())
```
petrkotas: There is better way of using the Expect. You can place annotation in the Expect. When it fails, the label is than shown and produce much clearer information of failure. For instance you code here will output:
```
Expected false to be true.
```
But when using the label `Expect(len(IOThreadVMI.Name) > symbols_to_compare).To(BeTrue(), "Should have some symbols...") will output
```
Should have some symbols...:
Expected false to be true.
```
which is much cleaner to read.

Please update the expects.

tests/iothreads_test.go
```diff
@@ -0,0 +1,206 @@
...
+				}
+			}
+			Expect(ded1_present).To(BeTrue())
+
+			// Step 5 is NA for automatic testing
+
+			duration := time.Duration(60)*time.Second
+			time.Sleep(duration)
+			//tests.WaitUntilVMIReadyWithNamespace(tests.NamespaceTestDefault, resultVMI, tests.LoggedInCirrosExpecter)
+			command := "/usr/local/bin/oc project kubevirt-test-default && "
```
petrkotas: In core, we have a helper function to accomplish this https://github.com/kubevirt/kubevirt/blob/6e6c1e42f2e78c3d6b0e8377a5550ada8ddb5a49/tests/utils.go#L1867

I believe you can build your own with this idea :).

sHaggYcaT: I seen it. I especially used another way to be sure that qemu run exactly what I check. Because It's test - I think bad idea use here implemented function.

gbenhaim:
`ERROR: It seems like you need to run 'make generate'. Please run it and commit the changes.`

tests/iothreads_test.go
petrkotas: I see. If you are really writing such manual cases, it would be much easier to use different approach, than golang.

tests/iothreads_test.go
ILpinto: the test is run with parameter from CLI [1],  so we need to parse them
[1]
${TESTS_OUT_DIR}/tests.test -kubeconfig=$kubeconfig -tag=$tag -prefix=$prefix -oc-path=${OC_PATH} -kubectl-path=${KUBECTL_PATH} -virtctl-path=${VIRTCTL_PATH} ${FUNC_TEST_ARGS}

tests/iothreads_test.go
sHaggYcaT: 7(about half of all) other tests already uses tests packages for several tasks (panic on error and cleanup). I've just their example

tests/iothreads_test.go
sHaggYcaT: I've added By statements instead of "Step" comments.

tests/iothreads_test.go
sHaggYcaT: I've added comments to all Expect statements except "Expect(err).ToNot(HaveOccurred())"

tests/iothreads_test.go
ILpinto: you you need it

tests/iothreads_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	BeforeEach(func() {
+		tests.BeforeTestCleanup()
+		IOThreadVMI = tests.NewRandomVMI()
+		IOThreadVMI.Spec = IOThreadSpec
+	})
+
+	Context("IOThreads Policies", func() {
+
+		// CNV-1046
```
ILpinto: remove it

tests/iothreads_test.go
ILpinto: Please use the method for getting the XML domain no need to rewrite code.

PR #496, opened 2018-11-29
Support SR-IOV deployment

booxter:

Fixes #

**Special notes for your reviewer**: this PR doesn't handle host configuration (BIOS, kernel, IOMMU, VF enablement). The reason is that it's not trivial to determine which PCI root devices to use to configure for SR-IOV DP plugin, neither it's trivial to autodetect VFs / configure kernel cmdline etc. There should be a different component (kubernetes-nmstate?) that will handle VF configuration pre-flight.

**Release note**:

```release-note
Support SR-IOV control plane deployment
```


booxter:
This is *not* ready yet (I can't validate the code works because of bugs in installer modules).

booxter:
I need to disable sriov deployment by default since it requires some pre-deployment host configuration for DP that is not automated and hence breaks the deployment. Otherwise, it is proved locally to work, so we could move to merge it after the default is flipped to false.

booxter:
@myakove I think CI fails because ansible book configures /etc/pcidp/config.json on localhost instead of on cluster nodes (it works locally for me because it's the same host). How do I deploy config files to cluster nodes? Is there something I can do in ansible to copy a generated file to remove hosts, or should I construct some k8s object that would run on nodes and create the right files?

booxter:
@rhatdan

Hi Dan, may I ask you to check if what I am doing here in terms of SELinux for SR-IOV device plugin deployed on Kubevirt cluster is sane enough? The relevant commit is here: https://github.com/kubevirt/kubevirt-ansible/pull/496/commits/b9305a891f89c25525ef520dda527b79b6ab148d

I've mimicked what we are already doing for another kubevirt container (virt-launcher) that runs libvirtd. While spc_t technically works, I wonder if we can do better here, for example, having a special type crafted specifically for kubernetes device plugins that all need to create Unix sockets under /var/lib/kubelet/device-plugins/ directory. I am not particularly well informed about best practices, and perhaps you already had experience with gardening other kubernetes device plugins that you could leverage here.

Thanks!

rthallisey:
+1 from me.  Needs a rebase

booxter:
@rthallisey thanks for merging. Is it regular practice to squash-merge PRs in this repo? I was trying to keep the branch clean and logically split, with clear commit messages, but now that it's squash-merged, these comments are more of a mess, and we can't later revert a single patch from the series. Should I continue building clean multi-patch series, or it doesn't matter and it will be squashed anyway?

rthallisey:
The policy is to squash and merge.

I appreciate that you take the time to write good commit messages; they're very helpful.  How about you pool them all together at to HEAD and have brief commit messages for the child commits?  Also, I think it makes sense to keep combining commits because my review capacity is pretty limited.  It helps for me to have it all in front of me :).

PR #499, opened 2018-12-03
add auto_IOThread_policy

sHaggYcaT:

It's test case for auto IOThread policy

Despite I calling BeforeTestCleanup my VMI and pod remains undeleted and this causes the fail of the second test.
Both tests work fine if they are run separately, but I decided not to split them into 2 different files since they are part of one test case.
If this is not a bug in BeforeTestCleanup I need to learn why my environment are not cleaned between tests.

### DIFF
--- a/tests/iothreads_auto_test.go
+++ b/tests/iothreads_auto_test.go
@@ -22,7 +22,6 @@ package tests_test
 import (
 	"flag"
 	"os/exec"
-	"time"
 	"encoding/xml"
 	"strings"
 
@@ -187,6 +186,8 @@ var _ = Describe("IOThreads", func() {
 			By("Creating VMI with desired spec")
 			IOThreadVMI, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Create(IOThreadVMI)
 			Expect(err).ToNot(HaveOccurred())
+			tests.WaitForSuccessfulVMIStart(IOThreadVMI)
+
 			Expect(len(IOThreadVMI.Name) > symbols_to_compare).To(BeTrue(), "VMI Name should contain at least N symbols")
 
 			By("Checking that corresponding pod exists")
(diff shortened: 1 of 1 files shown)


sHaggYcaT:
@petrkotas @ILpinto ?

tests/iothreads_auto_test.go
```diff
@@ -0,0 +1,334 @@
...
+			// and these symbols are different for pod name and vm name. So, I only can compare their names before
+			// these random symbols.
+			Expect(podList.Items[0].Name).To(HavePrefix("virt-launcher-" + IOThreadVMI.Name[:symbols_to_compare]), "Pod name should have a name similiar to VM name")
+
+			By("Checking that VMI with this name does exist")
+			getOptions := metav1.GetOptions{}
+			resultVMI, err := virtClient.VirtualMachineInstance(tests.NamespaceTestDefault).Get(IOThreadVMI.Name, &getOptions)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(*resultVMI.Spec.Domain.IOThreadsPolicy).To(Equal(autoPolicy), "Spec should have auto policy")
```
ILpinto: The create of VM with RandomVMI return VMI object.
on and with this object you can get all you need.
Like: VMI name:   vmi.name
You can run it with: ktests.WaitForSuccessfulVMIStart(vmi) 
And for the connect to contain you have the launcher pod  
pod that runs the VMI: 				
vmiPod := ktests.GetRunningPodByVirtualMachineInstance(vmi, tests.NamespaceTestDefault)

Please fix your code to use it.

sHaggYcaT: I've added these function, please check the updated PR. I removed delay that was there to wait for VMI to start (now WaitForSuccessfulVMIStart handles it), but I still receive list of PODs and check that only one POD is exist. If you think it is better to remove this check completely, and only check the name of the POD associated with current VMI - I can remove obtaining list of the pods.

PR #500, opened 2018-12-03
add templates test verification testcase

sHaggYcaT:
This test case checks if common-templates.yaml is valid for UI or not.

### DIFF
--- a/tests/templates_verification_test.go
+++ b/tests/templates_verification_test.go
@@ -30,6 +30,8 @@ import (
 	. "github.com/onsi/gomega"
 
 	"kubevirt.io/kubevirt/tests"
+        "os/exec"
+
 )
 
 func search_for_pattern(yaml_segment string, pattern string) bool {
@@ -56,7 +58,10 @@ var _ = Describe("Common templates", func() {
 
 			// Getting common_templates
 			// TODO: replace downloading common-templates with getting common-templates from RPM
-			ct_yml_url := "https://github.com/kubevirt/common-templates/releases/download/v0.3.1/common-templates-v0.3.1.yaml"
+
+                        ct_yml_url_byte, err := exec.Command("/bin/bash", "-c", "curl -s https://api.github.com/repos/kubevirt/common-templates/releases/latest | grep browser_download_url | cut -d '\"' -f 4").Output()
+                        ct_yml_url :=  string(ct_yml_url_byte)
+                        Expect(err).ToNot(HaveOccurred())
 			response, err := http.Get(ct_yml_url)
 			Expect(err).NotTo(HaveOccurred())
 			defer response.Body.Close()
--- a/tests/templates_verification_test.go
+++ b/tests/templates_verification_test.go
@@ -1,0 +1,82 @@
+/*
+ * This file is part of the KubeVirt project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


tests/templates_verification_test.go
```diff
@@ -0,0 +1,82 @@
...
+	})
+
+	Context("Testing generated templates", func() {
+
+		// CNV-1069
+		It("Check if template valid for UI", func() {
+
+			// Getting common_templates
+			// TODO: replace downloading common-templates with getting common-templates from RPM
+			ct_yml_url := "https://github.com/kubevirt/common-templates/releases/download/v0.3.1/common-templates-v0.3.1.yaml"
```
MarSik: Fixed version? What happens when we release a new one?

tests/templates_verification_test.go
```diff
@@ -0,0 +1,82 @@
...
+				By("Checking that template contains required lables")
+				Expect(search_for_pattern(yaml_segment, "\\s+os.template.cnv.io/[a-z0-9\\.]+:\\s\"true\"$")).To(BeTrue(), "Template should have os label")
+				Expect(search_for_pattern(yaml_segment, "\\s+workload.template.cnv.io/[a-z]+:\\s\"true\"$")).To(BeTrue(), "Template should have workload label")
+				Expect(search_for_pattern(yaml_segment, "\\s+flavor.template.cnv.io/[a-z]+:\\s\"true\"$")).To(BeTrue(), "Template should have flavor label")
+				Expect(search_for_pattern(yaml_segment, "\\s+template.cnv.io/type:\\s\"base\"$")).To(BeTrue(), "Template should have type base")
+
+			}
+		})
+	})
+})
```
MarSik: Is this it? No checks for the presence of major operating systems for example?

MarSik:
@cynepco3hahue @petrkotas can you please check this or add other QA reviewers? I am not sure who to add.

tests/templates_verification_test.go
```diff
@@ -0,0 +1,82 @@
...
+			common_templates_yaml := string(data)
+
+			for _, yaml_segment := range strings.Split(common_templates_yaml, "\n---\n") {
+
+				if !search_for_pattern(yaml_segment, "^Kind: Template$") {
+					continue
+				}
+
+				By("Checking that template contains required lables")
+				Expect(search_for_pattern(yaml_segment, "\\s+os.template.cnv.io/[a-z0-9\\.]+:\\s\"true\"$")).To(BeTrue(), "Template should have os label")
```
petrkotas: How does the UI figures out that the template is valid? I would like to see it referenced here and than strictly follow the mechanics in the test.

MarSik:
The UI doesn't do any validation. Openshift does when creating an instance. The labels and annotations are a convention that is described here https://github.com/kubevirt/common-templates/blob/master/templates/README.md

tests/templates_verification_test.go
sHaggYcaT: We have manual test plan. It was written based by recommendations from UI plan. I can share it you and can explain. Can we have some kind of video call or text chat or maybe email conversation?

petrkotas:
I think we are good :)
--
s pozdravem,
Petr Kotas

sHaggYcaT:
I've made rebase & fix (see description above)

tests/templates_verification_test.go
sHaggYcaT: I've implemented determination of the manifest's latest version (using github API)

tests/templates_verification_test.go
sHaggYcaT: I'm just following manual test plan that doesn't have any OS checks. @MarSik  What kind of check do you expect here?

tests/templates_verification_test.go
sHaggYcaT: @MarSik ^^

MarSik:
@sHaggYcaT I do not have merge rights and the tests are not passing..

MarSik:
ci please test

PR #526, opened 2018-12-24
Allowed to run the playbook on a remote server and not only on the master

SchSeba:
Allow remote execution of the playbook.

the first problem when you deploy it from a remove server is you may have only one cli (oc or kubectl) there is no need for both just depend on the type of the cluster, this is the first change I made.

The second change is to run this tests locally from the deployment machine and not from the master[0] as most of the tests do today.
The third change was on the kubevirt playbook to import other playbooks and not copy/paste them.
The last change was on the CI check-patch script to load the vars/all.yml as requested on the read me and not configure variables on every playbook. (-e@vars/all.yml)

**Release note**:

```release-note
Allow the admin to be able to deploy kubevirt on an existing cluster without the need to login into the master node and run it from there, just like openshift-ansbile.
Using the kubernetes API and having ssh connection to the masters and nodes.
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/kubevirt-ansible/526)


automation/check-patch.sh
```diff
@@ -209,7 +209,7 @@ run() {
         "ansible_modules_version=$ansible_modules_version"
         "kubevirt_openshift_version=$kubevirt_openshift_version"
         "openshift_playbook_path=$openshift_playbook_path"
-        "deploy_skydive=False"
+        "deploy_skydive=false"
```
dankenigsberg: is this relevant to the patch?

dankenigsberg:
The release note sounds frightening!
I'd rather replace it with something more verbose, e.g.
```
KubeVirt users should first have a Kubernetes (or OpenShift) cluster deployed, know where it is, and have admin credentials to it. Then they should be able to install kubevirt-ansible on their laptop, grant it access to the cluster, and execute the playbook to deploy KubeVirt. We must not ask them to ssh into one of their cluster nodes. Using the kAPI should be enough.
```
but probably not THAT verbose.

roles/cdi/templates/cdi-controller.yaml.j2
```diff
@@ -0,0 +1,322 @@
+kind: ServiceAccount
```
dankenigsberg: why do you introduce this? is it related to the patch?
I have a vague memory that we ship (and want to keep shipping) the cdi manifest elsewhere, which kubevirt-ansible.rpm requires.

dankenigsberg:
@cynepco3hahue I suspect your review would be better than mine

automation/check-patch.sh
SchSeba: Boolean in ansible with lower characters

roles/cdi/templates/cdi-controller.yaml.j2
SchSeba: I need to check I think the cdi download it and save it there in the playbook run time. I didn't add this file manually

automation/check-patch.sh
dankenigsberg: I understand - but think that it belongs to a separate and quick bugfix PR.

SchSeba: I also changed other parameters there so I think we can leave this change.

SchSeba:
ci test please

SchSeba:
@cynepco3hahue Can you please take a look on this PR?

cynepco3hahue:
I can see that previous implementation executes `kubectl` and `oc` commands to deploy kubevirt, so I do not see the problem to deploy it on the remote host without login to the master you just need to use correct `kubeconfig`, so can you please explain what problem did you have?

SchSeba:
Hi @cynepco3hahue so the first problem when you deploy it from a remove server is you may have only one cli (`oc` or `kubectl`) there is no need for both just depend on the type of the cluster, this is the first change I made.

The second change is to run this tests locally from the deployment machine and not from the master[0] as most of the tests do today. 
The third change was on the kubevirt playbook to import other playbooks and not copy/paste them.
The last change was on the CI `check-patch` script to load the `vars/all.yml` as requested on the read me and not configure variables on every playbook. (-e@vars/all.yml)

also if I remember correctly @myakove have some problem to deploy it from is local laptop into the cluster when he run it not as root maybe he can explain.

SchSeba:
> In this case you need to delete this vars from playbooks/cluster/openshift/config.yml

already loads the variables from this file
```yaml
    - name: include_vars
      include_vars:
        file: "{{ item }}"
      with_items:
        - "{{ playbook_dir }}/../../../vars/all.yml"
        - "{{ playbook_dir }}/vars/{{ kubevirt_openshift_version }}.yml"
```

cynepco3hahue:
> already loads the variables from this file
> 
> ```yaml
>     - name: include_vars
>       include_vars:
>         file: "{{ item }}"
>       with_items:
>         - "{{ playbook_dir }}/../../../vars/all.yml"
>         - "{{ playbook_dir }}/vars/{{ kubevirt_openshift_version }}.yml"
> ```

In this case you do not need to pass additiona `-e` parameter because under the CI it included via this playbook(I think it is the single place)

SchSeba:
> In this case you do not need to pass additiona -e parameter because under the CI it included via this playbook(I think it is the single place)

I already try it and it fails.
This is the change I made to make it work
https://github.com/kubevirt/kubevirt-ansible/pull/526/files#diff-8bd34cd4f8b2c2164d699e57d43b5b41R226

automation/check-patch.sh
```diff
@@ -205,12 +205,12 @@ run() {
     args+=(
         "provider=$provider"
         "inventory_file=$inventory_file"
-        "cluster=$cluster"
         "ansible_modules_version=$ansible_modules_version"
         "kubevirt_openshift_version=$kubevirt_openshift_version"
         "openshift_playbook_path=$openshift_playbook_path"
-        "deploy_skydive=False"
-	"storage_role=$storage_role"
+	    "storage_role=$storage_role"
```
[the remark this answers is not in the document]
SchSeba: Done

SchSeba:
@cynepco3hahue I make a rebase can you please take a look?

cynepco3hahue:
@SchSeba Looks like you re-introduced the problem with ns
```
Error from server (Forbidden): error when deleting "/tmp/kubevirt.yaml": namespaces "kube-system" is forbidden: this namespace may not be deleted
```
Can you please check it?

SchSeba:
@cynepco3hahue Can you please take a look now?

roles/cdi/tasks/provision.yml
```diff
@@ -1,30 +1,26 @@
...
-- name: Determine Environment
-  shell: "oc version> /dev/null 2>&1; if [ \"$?\" -eq 127 ]; then echo 'kubectl'; else echo 'oc'; fi"
-  register: cli
-
 - name: Check if namespace {{ cdi_namespace }} exists
-  shell: kubectl get ns | grep -w {{ cdi_namespace }} | awk '{ print $1 }'
+  shell: "{{ cluster_command }} get ns | grep -w {{ cdi_namespace }} | awk '{ print $1 }'"
   register: ns
 
 - name: Create {{ cdi_namespace }} namespace using kubectl
```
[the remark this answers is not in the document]
cynepco3hahue: I found this one, but I believe all points already less relevant.

SchSeba:
For reference https://jenkins.ovirt.org/job/kubevirt_kubevirt-ansible_standard-check-pr/2095/artifact/ci_build_summary.html

I don't think the problem is related to this PR I will rerun the test

SchSeba:
ci test please

roles/cdi/tasks/provision.yml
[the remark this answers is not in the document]
cynepco3hahue: @SchSeba I just now saw that I did not add the link https://www.reddit.com/r/openshift/comments/78jcpk/namespaces_vs_projects/
:smile:

PR #539, opened 2019-01-08
distribute sr-iov configuration using initContainer

phoracek:
In order to make network playbook operator friendly, we need to move
node configuration to daemon sets. With this patch, node sr-iov DP
configuration is stored in cluster-level config map and saved on a
node using SR-IOV DP init container.

The config map contains SR-IOV configuration per node (data key is
name of the node) and 'fallback-config' in case specific node configuration
is not provided.

In order to reconfigure node's SR-IOV conf, config map must be updated
and respective pod restarted.


phoracek:
@booxter please take a look

roles/network-multus/templates/cni-plugins.yml
```diff
@@ -1,4 +1,35 @@
...
+  name: sriov-nodes-config
+  namespace: {{ network_namespace }}
+data:
+  fallback-config: |
+    {
+        "resourceList":
+        [
+            {
+                "resourceName": "sriov",
+                "rootDevices": [],
```
booxter: We should expose the current variable with the list of root devices here, so that setting it configures all nodes to use it. (We should later rethink how we configure these devices per node but it can wait.)

roles/network-multus/templates/cni-plugins.yml
```diff
@@ -21,6 +52,39 @@ spec:
       - key: node-role.kubernetes.io/master
         operator: Exists
         effect: NoSchedule
+      initContainers:
+      - name: config-container
+        image: fedora:28
```
booxter: reason this is not 29?

phoracek: no, i just live in the past

roles/network-multus/templates/cni-plugins.yml
phoracek: Done

phoracek:
@booxter I fixed your comments

phoracek:
@nellyc @sHaggYcaT @myakove @cynepco3hahue hi, could you please take a look and merge if possible? Thanks

myakove:
LGTM

PR #542, opened 2019-01-10
Add kubevirt-ssp-operator deployment

yanirq:
Use Ansible to orchestrate kubevirt-ssp-operator deployment while OLM continues to develop.

**Release note**:

```release-note
Launch kubevirt-ssp-operator 
```

### DIFF
--- a/roles/kubevirt-ssp/defaults/main.yaml
+++ b/roles/kubevirt-ssp/defaults/main.yaml
@@ -1,0 +1,5 @@
+#TODO change default namespace?
+kubevirt_ssp_namespace: "{{ kubevirt_ssp_image_namespace | default('kube-system') }}"
+kubevirt_ssp_template_dir: "{{ role_path }}/templates"
+
+
--- a/roles/kubevirt-ssp/tasks/deprovision.yml
+++ b/roles/kubevirt-ssp/tasks/deprovision.yml
@@ -1,0 +1,18 @@
+---
+- name: Check that kubevirt-ssp-operator.yml still exists in /tmp
+  stat:
+    path: "/tmp/kubevirt-ssp-operator.yml"
+  register: kubevirt_ssp_template
+
+- name: Check for kubevirt-ssp-operator.yml.j2 template in {{ kubevirt_ssp_template_dir }}
+  stat:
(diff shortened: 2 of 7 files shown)


roles/kubevirt-ssp/defaults/main.yaml
```diff
@@ -0,0 +1,5 @@
+#TODO change default namespace?
+kubevirt_ssp_namespace: "{{ kubevirt_ssp_image_namespace | default('kube-system') }}"
```
MarSik: Shouldn't the default be kubevirt now?

roles/kubevirt-ssp/tasks/deprovision.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Check that kubevirt-ssp-operator.yml still exists in /tmp
+  stat:
```
MarSik: Why don't you use the https://docs.ansible.com/ansible/2.4/tempfile_module.html module?

MarSik: Is that because you want to avoid instantiating the template again when executed again?

yanirq: I was using the same convention as in other playbooks in the project, but I can change it accordingly

MarSik: Oh, lets keep it then, but I can't say I like this style..

roles/kubevirt-ssp/defaults/main.yaml
rthallisey: Should be kubevirt by default

roles/kubevirt-ssp/tasks/provision.yml
```diff
@@ -0,0 +1,20 @@
...
+
+- name: Create Kubevirt-SSP operator
+  command: "{{ cluster_command }} apply -f /tmp/kubevirt-ssp-operator.yml"
+
+- name: Create Kubevirt-SSP
+  block:
+  - name:
+  command: "{{ cluster_command }} apply -f /tmp/kubevirt_v1_kubevirtssp_crd.yaml"
+#  - name:
+#  command: "{{ cluster_command }} apply -f /tmp/kubevirt_v1_kubevirtssp_cr.yaml"
```
rthallisey: remove comment

PR #549, opened 2019-01-14
Add ansible role to deploy common templates

yanirq:
**What this PR does / why we need it:**
Role added for Kubevirt-ssp.
This role will deploy the common-templates file (taken from: https://github.com/kubevirt/common-templates/releases) 

**Special notes for your reviewer:**
once released, kubevirt-ssp-operator will replace this type of deployment.

**Release note**:

```release-note
Deploy common-templates in openshift enviroment.
```


yanirq:
@MarSik  @rthallisey

rthallisey:
Also needs some tasks for deprovision.

MarSik:
@rthallisey Do we still follow the APB structure then? @yanirq Here is your answer. I propose you query for Templatre objects in the proper namespace and with the proper labels and delete those.

roles/kubevirt-ssp/tasks/provision.yml
```diff
@@ -0,0 +1,13 @@
...
+  register: mktemp
+  changed_when: False
+
+- name: Copy common templates yaml to temp directory
+  copy:
+    src: "{{ kubevirt_ssp_files_dir }}/common-templates-v0.4.0.yaml"
+    dest: "{{ mktemp.stdout }}/common-templates.yaml"  
+
+- name: Create Kubevirt Template
+  shell: "{{ cluster_command }} apply -f {{ mktemp.stdout }}/common-templates.yaml -n {{ kubevirt_ssp_namespace }}"
```
MarSik: The default namespace for common templates should be `openshift`

rthallisey:
We don't need to follow the APB structure any more, but we want to be able to delete resources we created from the deploy tasks.

yanirq:
Changed the templates file to be taken from URL, add deprovision.

roles/kubevirt-ssp/tasks/deprovision.yml
```diff
@@ -0,0 +1,23 @@
...
+- name: Get Common Templates from {{ common_templates_manifest_url }}
+  block:
+  - name: Check for common-templates manifest in /tmp
+    stat:
+      path: "/tmp/common-templates.yaml"
+    register: downloaded_templates
+
+  - name: Download common-templates manifest
+    get_url:
+      url: "{{ common_templates_manifest_url }}/{{ release_tag }}/common-templates-{{ release_tag }}.yaml"
```
MarSik: What about downstream? Will this format work?

yanirq: the release_tag here is assigned in defaults (0.4.0 for example)

MarSik: Sure, but downstream download url might not match the base/release/common-templates-..yml format. And it should not download from upstream sources. So the role should support that case.

rthallisey: I don't like the idea of downloading the templates.  They caused issues for us all over the place.   For this release, can you either ship the template in kubevirt-ansible or add the templates as an rpm dependency of kubevirt-ansible?

yanirq:
Closing this PR in favour of a new one which will include the final format of the common templates file.

PR #582, opened 2019-01-24
Add sleep container in cpu-node-labeller to prevent  restarting

ksimon1:
Moved cpu-node-labeller into initContainers. Added second cpu-node-labeller into containers to sleep infinity and to hold pod running (so it is not restarted again and again), because when daemonset pod finishes, it is restarted everytime it finishes (restartPolicy can't be turned off for daemonset pods).

**Release note**:
```release-note
NONE
```

### DIFF
--- a/roles/kubevirt-cpu-node-labeller/templates/kubevirt-cpu-node-labeller-0.0.1.yaml
+++ b/roles/kubevirt-cpu-node-labeller/templates/kubevirt-cpu-node-labeller-0.0.1.yaml
@@ -56,16 +56,10 @@ spec:
     spec:
       serviceAccount: kubevirt-cpu-node-labeller
       containers:
-        - env:
-          - name: NODE_NAME
-            valueFrom:
-              fieldRef:
-                fieldPath: spec.nodeName
-          image: {{ docker_prefix }}/kubevirt-cpu-node-labeller:{{ docker_tag }}
-          name: kubevirt-cpu-node-labeller
-          volumeMounts:
-            - name: nfd-source
-              mountPath: "/etc/kubernetes/node-feature-discovery/source.d/"
+      - name: kubevirt-cpu-node-labeller-sleeper
+        image: {{ docker_prefix }}/kubevirt-cpu-node-labeller:{{ docker_tag }}
+        command: ["sleep"]
+        args: ["infinity"]
       initContainers:
         - image: {{ docker_prefix }}/kubevirt-cpu-model-nfd-plugin:{{ docker_tag }}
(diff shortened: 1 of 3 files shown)


roles/kubevirt-cpu-node-labeller/templates/kubevirt-cpu-node-labeller-0.0.1.yaml
```diff
@@ -56,16 +56,10 @@ spec:
...
-              fieldRef:
-                fieldPath: spec.nodeName
-          image: {{ docker_prefix }}/kubevirt-cpu-node-labeller:{{ docker_tag }}
-          name: kubevirt-cpu-node-labeller
-          volumeMounts:
-            - name: nfd-source
-              mountPath: "/etc/kubernetes/node-feature-discovery/source.d/"
+      - name: kubevirt-cpu-node-labeller-sleeper
+        image: {{ docker_prefix }}/kubevirt-cpu-node-labeller:{{ docker_tag }}
+        command: ["sleep"]
```
rthallisey: What's the purpose of the container if it's sleeping?

MarSik: Daemonset has to be up and running or kubernetes tries to restart it. But we only need one time action for every node that populates the cpu model labels. That is why we had to add an always on sleeping container to prevent kubernetes from going into crashed state and restart.

ksimon1: it keeps pod running, so it is not restarted. In tests, the cpu-node-labeller sucessfully finished, but openshift restarts the pod, which in some cases ends with CrashLoopBackOff state. Daemonset pods has to be running all the time, otherwise they are restarted.

MarSik: Btw this seems to be the common workaround for run-once DaemonSets per https://github.com/kubernetes/kubernetes/issues/50689 and https://github.com/kubernetes/kubernetes/issues/36601

rthallisey: I have a feeling kubernetes has an object or configuration setting that will solve this problem better.  We can reevaluate later.

rthallisey:
Same test failure as https://github.com/kubevirt/kubevirt-ansible/pull/581 so this PR isn't breaking CI.

PR #592, opened 2019-02-01
Deprovision sriov resources

booxter:
**What this PR does / why we need it**: delete SR-IOV resources on deprovisioning. The only piece that is not cleaned up is /etc/pcidp because I don't know how to do it easily without running a new container that would delete the directory, on all cluster nodes, which seems like overkill.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/roles/network-multus/tasks/deprovision.yml
+++ b/roles/network-multus/tasks/deprovision.yml
@@ -25,6 +25,40 @@
 - name: Delete cni plugins Resources
   command: "{{ cluster_command }} delete -f /tmp/cni-plugins.yml --ignore-not-found"
 
+# note: this rule does NOT delete /etc/pcidp/config.json file because k8s
+# doesn't support finalizers to clean up mounted directory
+- name: Uninstall SR-IOV device plugin
+  block:
+  - name: Render SR-IOV DP deployment yaml
+    template:
+      src: sriovdp.yml
+      dest: /tmp/sriovdp.yml
+
+  - name: Delete SR-IOV DP resources
+    shell: "{{ cluster_command }} delete -f /tmp/sriovdp.yml --ignore-not-found"
+
+- name: Uninstall SR-IOV CNI plugin
+  block:
+  - name: Render SR-IOV CNI deployment yaml
+    template:
+      src: sriov-cni.yml
+      dest: /tmp/sriov-cni.yml
+
+  - name: Delete SR-IOV CNI resources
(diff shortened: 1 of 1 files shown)


booxter:
@rthallisey @phoracek @SchSeba could you please review? Thanks.

roles/network-multus/tasks/deprovision.yml
```diff
@@ -25,6 +25,40 @@
...
   command: "{{ cluster_command }} delete -f /tmp/cni-plugins.yml --ignore-not-found"
 
+# note: this rule does NOT delete /etc/pcidp/config.json file because k8s
+# doesn't support finalizers to clean up mounted directory
+- name: Uninstall SR-IOV device plugin
+  block:
+  - name: Render SR-IOV DP deployment yaml
+    template:
+      src: sriovdp.yml
+      dest: /tmp/sriovdp.yml
```
rthallisey: should this only be run when sriov is enabled?

SchSeba: +1  @booxter we can wrap everything in a block. and check for the sriov boolean

roles/network-multus/tasks/deprovision.yml
```diff
@@ -25,6 +25,40 @@
...
+    template:
+      src: sriov-cni.yml
+      dest: /tmp/sriov-cni.yml
+  - name: Delete SR-IOV CNI resources
+    shell: "{{ cluster_command }} delete -f /tmp/sriov-cni.yml --ignore-not-found"
+  when: deploy_sriov_plugin
+
+- name: Uninstall SR-IOV network
+  block:
+  - name: Render SR-IOV network CRD yaml
```
phoracek: Just curious: Why do we deploy an SR-IOV network attachment definition ourselves?

booxter: This is not a concern to address in this PR, but in general, the idea is that VMIs can be attached to SR-IOV right after deployment. The CRD, in case of SR-IOV, is to reflect pool name that is configured in config.json for DP, which is hardcoded right now. I find it logical that if we configure DP then we should also configure multus network to consume the resource. Again, this is a separate discussion from this PR.

rthallisey:
@booxter if we don't cleanup /etc/pcidp on all nodes will it be an issue on redeploy?

booxter:
@rthallisey it shouldn't be an issue on redeploy. We override an existing file via configmap.

PR #601, opened 2019-02-14
Tests add multus bond

myakove:
Test connectivity between 2 VMs when VM interface is on top of OVS bridge with BOND port.

```release-note
NONE
```

### DIFF
--- a/tests/network_connectivity_test.go
+++ b/tests/network_connectivity_test.go
@@ -20,9 +20,10 @@ import (
 
 const (
 	privilegedPodYaml  = "tests/manifests/privileged-pod.yml"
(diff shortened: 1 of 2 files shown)


tests/network_connectivity_test.go
```diff
@@ -219,6 +321,21 @@ var _ = Describe("[rfe_id:694][crit:medium][vendor:cnv-qe@redhat.com][level:comp
 		}, 30*time.Second)
 		Expect(err).ToNot(HaveOccurred())
 	})
+	It("[test_id:1719]Connectivity between VM and VM - Multus, OVS on BOND", func() {
+		if !bondSupportedNodes {
+			Skip("Nodes are not support BOND")
```
yossisegev: Nodes DO not support BOND

PR #604, opened 2019-02-27
Add networkpolicy test:Set connectivity to only allow from http

duyanyan:
Add networkpolicy test:Set connectivity to only allow from http
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/networkpolicy_test.go
+++ b/tests/networkpolicy_test.go
@@ -1,0 +1,110 @@
+package tests_test
+
+import (
+	"flag"
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	expect "github.com/google/goexpect"
+
+	k8sv1 "k8s.io/api/core/v1"
+	v13 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	tests "kubevirt.io/kubevirt-ansible/tests/framework"
+	v1 "kubevirt.io/kubevirt/pkg/api/v1"
+	ktests "kubevirt.io/kubevirt/tests"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+	"kubevirt.io/kubevirt/pkg/kubecli"
+)
+
+const (
+	denyallYaml   = "tests/manifests/networkpolicy-deny-all.yml"
+	allowhttpYaml = "tests/manifests/networkpolicy-allow-http.yml"
+)
+
+var _ = Describe("[rfe_id:150][crit:high][vendor:cnv-qe@redhat.com][level:component]Networkpolicy", func() {
+	flag.Parse()
+
+	virtClient, err := kubecli.GetKubevirtClient()
+	ktests.PanicOnError(err)
+
+	var vmia *v1.VirtualMachineInstance
+	var vmib *v1.VirtualMachineInstance
+
+	ktests.BeforeAll(func() {
+		// Flannel networking model doesn't support network policy
+		ktests.SkipIfUseFlannel(virtClient)
+		ktests.SkipIfNotUseNetworkPolicy(virtClient)
+		ktests.BeforeTestCleanup()
+	})
+
+	Context("Networkpolicy allow http port", func() {
+		It("[test_id:CNV-369] Set connectivity to only allow from http port", func() {
+
+			By("Create VMIs")
+			userData := "#cloud-config\npassword: fedora\nchpasswd: { expire: False }\n"
+			vmia = ktests.NewRandomVMIWithEphemeralDiskAndUserdata(ktests.ContainerDiskFor(ktests.ContainerDiskFedora), userData)
+			ktests.AddExplicitPodNetworkInterface(vmia)
+
+			vmia.Spec.Domain.Resources.Requests[k8sv1.ResourceName("memory")] = resource.MustParse("1024M")
+			_, err = virtClient.VirtualMachineInstance(ktests.NamespaceTestDefault).Create(vmia)
+			Expect(err).ToNot(HaveOccurred())
+
+			vmib = ktests.NewRandomVMIWithEphemeralDiskAndUserdata(ktests.ContainerDiskFor(ktests.ContainerDiskCirros), "#!/bin/bash\necho 'hello'\n")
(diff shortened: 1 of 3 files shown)


duyanyan:
@myakove @yossisegev Could you please help review?

tests/networkpolicy_test.go
```diff
@@ -0,0 +1,90 @@
...
+	flag.Parse()
+
+	virtClient, err := kubecli.GetKubevirtClient()
+	ktests.PanicOnError(err)
+
+	var vmia *v1.VirtualMachineInstance
+	var vmib *v1.VirtualMachineInstance
+
+	ktests.BeforeAll(func() {
+		ktests.SkipIfUseFlannel(virtClient)
```
yossisegev: I would add a documentation comment, to explain why Flannel networking model should be skipped, because as opposed to *SkipIfNotUseNetwotkPolicy* - this is not self-explanatory.
Something brief and simple, like:
`// Flannel networking model doesn't support network policy`
(I had to conclude that myself from [https://github.com/coreos/flannel#networking-details](url))

tests/networkpolicy_test.go
```diff
@@ -0,0 +1,90 @@
...
+			vmia, err = virtClient.VirtualMachineInstance(ktests.NamespaceTestDefault).Get(vmia.Name, &v13.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			ip := vmia.Status.Interfaces[0].IP
+			ktests.WaitUntilVMIReady(vmib, ktests.LoggedInFedoraExpecter)
+			err = ktests.CheckForTextExpecter(vmib, []expect.Batcher{
+				&expect.BSnd{S: fmt.Sprintf("curl %s:80 \n", ip)},
+				&expect.BExp{R: "test"},
+			}, 60)
+			Expect(err).ToNot(HaveOccurred())
+		})
```
yossisegev: The test only covers the "allow-tcp" policy, and its result would be the same even if you didn't
 apply the "deny-by-default" policy.
I think you should add a negative step here - try to connect to a port other than the allowed port (80), and expect it to fail.

duyanyan: Thanks for review. Yes, I was planning to do that, but I think the deny-all policy already covered by https://github.com/kubevirt/kubevirt/blob/master/tests/networkpolicy_test.go#L85

tests/networkpolicy_test.go
duyanyan: Added a comment in the step.

tests/networkpolicy_test.go
yossisegev: In the test you mentioned, the "deny-by-default" is covered as a standalone. The new test you composed should verify the  "deny-by-default" can co-exist with the "allow-tcp" policy. This is what is done in the [Polarion test](https://polarion.engineering.redhat.com/polarion/#/project/CNV/workitem?id=CNV-369) which is the source for this automation.
In your new test, you do apply the "deny-by-default" policy, so I think it is preferred to verify it is applied as expected, by trying to connect to a port !=80, and expect this connaction attempt to fail.

myakove: I agree with Yossi here, I think we should make sure that only port 80 is opened and we cannot access different port.

tests/networkpolicy_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}, 60)
+	Expect(err).ToNot(HaveOccurred())
+}
+
+
+func assertCurlSucceed(ip string, vmi *v1.VirtualMachineInstance) {
+	ktests.WaitUntilVMIReady(vmi, ktests.LoggedInCirrosExpecter)
+	err := ktests.CheckForTextExpecter(vmi, []expect.Batcher{
+		&expect.BSnd{S: fmt.Sprintf("curl --silent --connect-timeout 5 %s:80 | grep hellokubevirt | wc -l \n", ip)},
+		&expect.BExp{R: "1"},
```
myakove: Please make only one function that check curl and the function will get the expected result that you need.
expect.BExp{R: "1"},

these two functions seem duplicated.

duyanyan:
I have updated the code according all the comments. Thanks

tests/networkpolicy_test.go
duyanyan: Already updated. Thanks

yossisegev:
LGTM

myakove:
ci test please

myakove:
ci test please

duyanyan:
ci test please

yossisegev:
@duyanyan Why is this PR closed?
Isn't it going to be merged?

duyanyan:
ci test please

duyanyan:
ci test please

myakove:
ci test please

duyanyan:
The CI failed to setup test env due to the issue https://github.com/kubevirt/kubevirt-ansible/issues/613
But the code is passed in local, please check the log: https://paste.fedoraproject.org/paste/1KGvV4yyZy-gk3G2TuZ6zA
Could someone kindly help to merge it manually?

myakove:
@gbenhaim maybe remove the required check (openshift 3.11 is less relevant now) and we need to be able to merge patches.

tests/networkpolicy_test.go
SchSeba: Why we need both `ktests.SkipIfUseFlannel` and `ktests.SkipIfNotUseNetworkPolicy` ?

duyanyan: Because networkpolicy is not supported when using flannel or openshift other networkplugin, eg, redhat/openshift-ovs-subnet or redhat/openshift-ovs-multutenant. It only works on redhat/openshift-ovs-networkpolicy.

We could check the current on cluster: 
```
# oc get clusternetwork
NAME      CLUSTER NETWORK   SERVICE NETWORK   PLUGIN NAME
default   10.128.0.0/14     172.30.0.0/16     redhat/openshift-ovs-networkpolicy
```

SchSeba: ok so you are going to test it on k8s with other network cni plugins?

duyanyan: Actually OpenShift can use flannel as well, but seems we don't setup OCP + flannel env for our testing. Do you want me to remove it?

